PR #70, opened 2019-01-21
intro as a GIF

bictorman:
convert asciinema intro to GIF for readme


fho:
@bictorman I think it's better to keep the link.
Using a gif means that you have to load ~0.5MB when viewing the main page.
Also the font from the gif is hard to read and doesn't look good.

bictorman:
If I make it smaller looks like the text get's cut off, we would need to redo the asciicinema in a smaller terminal window.

intro

PR #74, opened 2019-04-11
Exec refactor

fho:
```
Author: Fabian Holler
Committer: Fabian Holler

        resolve: do not run "git ls-files" in a shell

        "git ls-files" was previously run in a shell. This caused that the paths
        that were passed to "git ls-files" were resolved by the shell to
        filenames. When the application directory contained a file that was not
        tracked in Git and the path for GitFile BuildInput was a glob that
        matched this file, resolving the build inputs failed.

        It also has the effect that the paths were matched against the files in
        the filesystem instead of against the tracked git files.

Author: Fabian Holler
Committer: Fabian Holler

        build: prefix console output with app name in verbose mode

        When build is done with --verbose enabled, the command output of an
        application is now prefixed with it's application name in yellow color.

Author: Fabian Holler
Committer: Fabian Holler

        exec: refactor package

        The exec package is refactored completely. Commands are run by creating
        a Cmd struct and calling it's Run() method. The Run() can be configured
        by calling methods on the Cmd() struct that can be chained like:
        ShellCommand("ls -1").Directory("/").ExpectSuccess().Run()

        Run() now returns an ExitCodeError() if ExpectSuccess() was set for the
        command. This simplifies having a consistent error message for all
        command executions and callers only have to check for the error instead
        of also doing an additional exitcode check.

        Packages that were using exec were adapted. Previously all commands were 
        executed in a shell. Now only the build command and git ls-files is run
        in a shell. git ls-files will be changed to also run without a shell in
        a following PR.

```


exec/command.go
```diff
@@ -0,0 +1,186 @@
...
+// Run executes the command.
+func (c *Cmd) Run() (*Result, error) {
+	cmd := exec.Command(c.path, c.args...)
+	cmd.Dir = c.dir
+
+	outReader, err := cmd.StdoutPipe()
+	if err != nil {
+		return nil, err
+	}
+	cmd.Stderr = cmd.Stdout
```
FranciscoKurpiel: For this PR I think this is ok. In general I think joining stdout and -err is too restrictive.

exec/command.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil, err
+	}
+
+	var exitCode int
+	err = cmd.Wait()
+	if err != nil {
+		if ee, ok := err.(*exec.ExitError); ok {
+			if status, ok := ee.Sys().(syscall.WaitStatus); ok {
+				exitCode = status.ExitStatus()
+			}
```
FranciscoKurpiel: I feel like there should be an *else* statement here

resolve/gitpath/gitpaths.go
```diff
@@ -28,8 +28,7 @@ func NewResolver(path string, globs ...string) *Resolver {
 // Resolve the glob paths to absolute file paths by calling
 // git ls-files
 func (r *Resolver) Resolve() ([]string, error) {
-	arg := strings.Join(r.globs, " ")
-	out, err := git.LsFiles(r.path, arg)
+	out, err := git.LsFiles(r.path, r.globs...)
```
FranciscoKurpiel: awesome

exec/command.go
```diff
@@ -0,0 +1,186 @@
...
+		args:         arg,
+		dir:          ".",
+		debugfFn:     DefaultDebugfFn,
+		debugfPrefix: DefaultDebugPrefix,
+	}
+}
+
+// ShellCommand executes a command in sh shell.
+func ShellCommand(cmd string) *Cmd {
+	return Command("sh", "-c", cmd)
```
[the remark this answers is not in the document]
fho: For baur's purpose I don't know a reason why somebody would want to use a different shell.

`ShellCommand` is only used for the build command to not having to either:
-  parse and convert the arguments in the build-command string to an args slice or 
- have the user to define them in an array in the config instead of one simple string.

Maybe there is an easy way to ensure correct argument handling and having not to run the command in a shell instead, that I'm not aware of? :-)

From the package's view and re-usability I agree that it would be nice to have it configurable.

exec/command.go
fho: Yes, maybe we will need in the future to distinguish between stdout and stderr.

exec/command.go
fho:
```
else {
    exitCode = 0
}
```
?
I think it only adds unnecessary lines. :-)

exec/command.go
```diff
@@ -0,0 +1,186 @@
...
+	DefaultDebugPrefix = "exec: "
+)
+
+// ExitCodeError is returned from Run() when a command exited with a code != 0.
+type ExitCodeError struct {
+	*Result
+}
+
+// Error returns the error description.
+func (e ExitCodeError) Error() string {
```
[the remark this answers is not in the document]
FranciscoKurpiel: Yeah, I think it makes sense, the DefaultDebugPrefix is intended to be used with the debug function, and this is not the case. Let's keep it as you initially wrote.

exec/command.go
```diff
@@ -0,0 +1,186 @@
...
+	debugfPrefix  string
+	expectSuccess bool
+}
+
+// Command returns a new Cmd struct.
+// If name contains no path separators, Command uses LookPath to
+// resolve name to a complete path if possible. Otherwise it uses name directly
+// as Path.
+// By default a command is run in the current working directory.
+func Command(name string, arg ...string) *Cmd {
```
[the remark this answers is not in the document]
FranciscoKurpiel: Yes, it's longer. I think it makes sense to keep the convention already used on the exec package.

exec/command.go
```diff
@@ -0,0 +1,186 @@
...
+		path:         name,
+		args:         arg,
+		dir:          ".",
+		debugfFn:     DefaultDebugfFn,
+		debugfPrefix: DefaultDebugPrefix,
+	}
+}
+
+// ShellCommand executes a command in sh shell.
+func ShellCommand(cmd string) *Cmd {
```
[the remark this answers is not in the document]
FranciscoKurpiel: Ok

exec/command.go
FranciscoKurpiel: We have three options:
1. The shortest code: `exitCode = ee.Sys().(syscall.WaitStatus).ExitStatus()`
2. Test type and ignore if type is not syscall.WaitStatus
3. Test type and return error if type don't match

Option 2 will silence a type conversion failure and assume that exitCode is 0. In my opinion the other two options are better.

fho: thanks a lot!
I fixed it like discussed verbally

PR #79, opened 2019-04-29
cfg: make the [Build] section in the .app.toml file optional

fho:
.app.toml files are now also valid if they have no [Build] section defined.
This allows to use baur for discovering apps in a repository without building
them with baur.

"baur ls apps" shows the status "Build Command Undefined" for apps that have no
[Build] section.
When "baur build" is called, apps without a [Build] section are skipped.

The "baur build" output is now more similar with and without "--force" option.
In both cases it now lists the applications that will be build before starting
the build.
If --force is passed, apps that will be build will be shown with Pending status.

The application list that is printed by "baur build" is now also sorted.

### DIFF
--- a/command/build.go
+++ b/command/build.go
@@ -360,6 +360,26 @@ func maxAppNameLen(apps []*baur.App) int {
 	return maxLen
 }
 
+func appsWithBuildCommand(apps []*baur.App) []*baur.App {
+	res := make([]*baur.App, 0, len(apps))
+
+	maxAppNameLen := maxAppNameLen(apps) + 4
+
+	for _, app := range apps {
+		if len(app.BuildCmd) == 0 {
+			fmt.Printf("%-*s => %s\n",
+				maxAppNameLen, app.Name, coloredBuildStatus(baur.BuildStatusBuildCommandUndefined))
+			continue
+		}
+
+		fmt.Printf("%-*s => %s\n",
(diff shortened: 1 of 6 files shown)


command/build.go
```diff
@@ -360,6 +360,26 @@ func maxAppNameLen(apps []*baur.App) int {
 	return maxLen
 }
 
+func appsWithBuildCommand(apps []*baur.App) []*baur.App {
+	res := make([]*baur.App, 0, len(apps))
+
+	maxAppNameLen := maxAppNameLen(apps) + 4
```
FranciscoKurpiel: - shadowing the package function with a local variable;
- What's the `+ 4`? It must be a const or at least have a relevant comment.

command/build.go
```diff
@@ -392,29 +414,41 @@ func buildRun(cmd *cobra.Command, args []string) {
...
+		fmt.Println("Outputs are not uploaded.")
+	}
+
 	if len(apps) == 0 {
-		fmt.Println()
 		term.PrintSep()
-		fmt.Println("Application build(s) already exist, nothing to build.\n" +
-			"If you want to rebuild applications pass '-f' to 'baur build'")
+
+		if !buildForce && len(apps) > 0 {
```
FranciscoKurpiel: `len(apps)` is always `0` because of the outer `if` statement

fho: thanks, fixed

command/build.go
fho: I improved both :-)

PR #90, opened 2019-11-11
ci: replace gometalinter with golangci-lint

fho:
Gometalinter was discontinued, golangci-lint is the tool to go.

- Switch to golangci-lint, enable some more linters on the way
- Fix the issues that golangci-lint found


FranciscoKurpiel:
@fho labels, plz :)

PR #93, opened 2019-12-26
fix editorconfig since `tabs` is wrong need to use `tab`

oneslash:
I was trying to work on this and noticed that my tabs are actually spaces, and decided to fix it.


fho:
@oneslash thanks!

Issue #96, opened 2020-02-03
Configuring .app.toml to use GCR Docker Repositories as a Docker Registry doesn't work 


nicodoggie:
In our environment, we use Google Cloud Repository to house our docker images, and the URI path does not necessarily conform to the assumptions made in the `parseRepositoryURI` function.

The code assumes a URI in the format `[<host[:port]>]/<owner>/<repository>:<tag>`, but in GCR the paths are much more flexible.

In our case we have something like `gcr.io/department/service-type/application:tag` instead.

Code in question:
https://github.com/simplesurance/baur/blob/cc5c3b13dae0671c0660689601111764b16baca7/upload/docker/docker.go#L150-L158

I get that the assumption conforms to `hub.docker.com`'s URI format, but it makes baur unable to support private docker repositories that do not conform to this format. 

Then, event if I conform to the format enforced by that code, it seems that `~/.docker/config.json` file isn't being respected anyway, whether or not the `DOCKER_CONFIG` environment variable is set. I haven't investigated that yet, however.

fho:
@nicodoggie thanks for the report.

Are you using the `Domain-scoped projects` syntax (https://cloud.google.com/container-registry/docs/overview#domain-scoped_projects)?
According to their documentation the syntax in deprecated.

nicodoggie:
No, as I described, we scope things by department and service type
Something like:
`gcr.io/ops-team/scheduled/runner:deadbeef12341234`

fho:
@nicodoggie that looks like the `Domain-scoped projects´ syntax.
If it's not, can you please provide a link to the GCR or better docker documentation that describes this syntax?
thanks

nicodoggie:
I don't really see it documented anywhere, but GCR does allow to push into arbitrarily long paths into gcr.io

`docker push gcr.io/project-id/service-type/repo:tag`
`docker push gcr.io/project-id/service-type/whatever/repo:tag`

is actually acceptable, and is used across the board in our infrastructure. I've added a PR for a proposed fix.

fho:
@nicodoggie 
Thanks for the PR.

I still would like to have a reference for this syntax. It's not mentioned in documentation for docker.

Why do you think it is not the https://cloud.google.com/container-registry/docs/overview#domain-scoped_projects syntax?

`[HOSTNAME]/[DOMAIN]/[PROJECT]/[IMAGE]`
matches
`gcr.io/project-id/service-type/repo`
[...]

PR #97, opened 2020-02-03
Fix parseRepositoryURI to accept URIs formatted as gcr.io/a/b/c/...

nicodoggie:
The fix was done by using `strings.SplitN` and limit the splits to 3. 

This is in attempt to fix https://github.com/simplesurance/baur/issues/96

### DIFF
--- a/upload/docker/docker_test.go
+++ b/upload/docker/docker_test.go
@@ -36,7 +36,13 @@ func Test_parseRepositoryURI(t *testing.T) {
 			wantRepository: "simplesurance/calculator",
 			wantTag:        "latest",
 		},
-
+		{
+			name:           "serverWithPortLongRepo",
+			destArg:        "localhost/simplesurance/app/calculator:latest",
+			wantServer:     "localhost",
+			wantRepository: "simplesurance/app/calculator",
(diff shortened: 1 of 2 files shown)


upload/docker/docker_test.go
```diff
@@ -36,7 +36,13 @@ func Test_parseRepositoryURI(t *testing.T) {
 			wantRepository: "simplesurance/calculator",
 			wantTag:        "latest",
 		},
-
```
fho: Having an empty server part and a repository name with multiple path components does not work.
The part before the first `/` is now misinterpreted as the server addr.

```
                {    
                        name:           "noServerLongRepo",                                                                                                                                                                                                    
                        destArg:        "simplesurance/app/calculator:latest",
                        wantServer:     "",
                        wantRepository: "simplesurance/app/calculator",
                        wantTag:        "latest",
                },
```

fho:
@nicodoggie See https://github.com/simplesurance/baur/pull/99
I fixed the issue by introducing a new config parameter for the .app.toml file to specify the docker registry.
The repository name can now contain multiple slashes.

PR #99, opened 2020-02-03
cfg: support to specify registry address for docker uploads, support slashes in docker repository names

fho:
```
        cfg: support specifying docker registry and repositories with multiple slashes
        
        - The DockerImageRegistryUpload section in the .app.toml file has a new field
          called "registry" to allow specifying the registry to that the image is
          uploaded.
          Specifying the registry address in the "repository" field in the config does
          not work anymore.
        - Repository names with multiple path components separated with "/" are now
          supported
          (https://github.com/docker/distribution/blob/master/docs/spec/api.md#overview)

```
        
This fixes: https://github.com/simplesurance/baur/issues/96


fho:
@nicodoggie would be great if you could give it a try and report if it works for you

nicodoggie:
Thanks! I'll give it a go!

PR #110, opened 2020-02-28
Fix docker authentication

fho:
The credentials for a docker registry were not found if:
- the auth configuration contained entries for multiple registries and
- the specified docker registry host was not an URL

These issues are fixed, see commit messages for details


fho:
merged manually

PR #118, opened 2020-03-26
Add branch awareness and previous build option

### LINKED ISSUE #117: Support per branch builds and previous build comparison
Hi,

First off I would just like to say thanks for this project, we've found it really useful.  We currently have a fork that we are using for our specific use case, and are wondering if you would accept a PR to support this.  Our use case is as follows:

We currently use Jenkins, Artifactory, and Spinnaker to support a CI/CD work flow.  Every time the code changes (even if it is reverting previous change) we want to produce a new artifact that we can then pass through our CD pipelines.  The versioning uses semver with the short commit hash as metadata.  To support this we have introduced a config option in to baur to just look at the last build when comparing the digests.

We also use the CI pipeline to build and deploy pull requests, and we would like these to be tracked separately to the mainline build.  This means that we trigger a build when the PR gets merged to master as the digests are tracked separately

dbason:
Fixes #117 
@fho here are the code changes we have made to support our use case as outlined in the issue.

### DIFF
--- a/app.go
+++ b/app.go
@@ -13,9 +13,10 @@ import (
 
 // App represents an application
 type App struct {
-	RelPath string
-	Path    string
-	Name    string
+	RelPath      string
+	Path         string
+	Name         string
+	UseLastBuild bool
 
 	repositoryRootPath string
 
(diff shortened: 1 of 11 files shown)


fho:
see https://github.com/simplesurance/baur/issues/117#issuecomment-614654001

Issue #78, opened 2019-04-29
3. Support to run arbitrary commands as tasks


fho:
Instead of only being able to define a `Build.Command` that can be run and tracked with baur, support to run and track arbitrary commands as tasks.
See https://github.com/simplesurance/baur/wiki/Tasks-Design for more information.

- Config:
    - [x] Adapt `.app.toml` cofiguration file syntax
   - [x] Replace Build section with Tasks slice in app config file, support to only have 1 task named "build"
   - [x] Handle merging of app config files with included file in cfg package
   - [x]  Include by include-ids instead of by filename
  -  ~~[ ] Support to define include section in app config files~~
  -  [x] Support to include inputs/outputs in Task-Includes
  - [x] Support to define Tasks in includes and to include them
  - [x] Support multiple Tasks per Application
 -  [x] implement merging `cfg.Tasks.Includes` with it's includes
 - [x] Review and improve comments in example configuration files, use examples for the includes that differ from the app config sections 
- Commandline interface
 - [x] run
  - [x] build (also rename to `run`)
  - [x] ls apps
  - [x] ls builds (became `ls runs`)
  - [x]  ls inputs
  - [x]  ls outputs
  - [x]  show
  - [x]  verify (command was removed
- Migration
  - ~~implement command to migrate database from old to new schema(?)~~
  - [x] implement command to migrate configs from old to new format
- Tests
  - [ ] `baur run` testcases
    - [X] App without outputs
    - [X] App with outputs
    - [ ] App without task
    - [ ] Task without inputs
- [x] Validate in cfg package App and Task Names, ensure they do not contain a '#' and '*', are > X characters long. Should we limit the names to letters, digits and punctuation chars? 
- [x] Adapt database layout
- [x] Document the supported variables in the config file somewhere
- [x] Update Example Repository

See https://github.com/simplesurance/baur/wiki/Tasks-Design for details

PR #155, opened 2020-06-17
Tasks: Implement run, show, ls runs commands

fho:
The PR adapts multiple commands to be able to work with Tasks instead of Builds.

- "baur build" is renamed to "baur run" and now supports to run tasks. A lot of related code was refactored.
Recording result of failed tasks will be implemented in a follow-up PR. The behavior regarding failed task runs and upload is the same then before, it cancels the whole run.
- "baur show" supports to show task information for an app and recorded task runs,
- "baur ls builds" was renamed to "baur ls runs", allow filtering for specific tasks and showing task information,
- command helper functions related to terminal output are moved to an own package

Related: https://github.com/simplesurance/baur/issues/78

------

I wonder if I should introduce a Baur struct in the future that could act as central API endpoint.
It could have methods for the functionality that is currently e.g. in `output.go`,  `uploader.go`, `storage.go`. 
Having a central struct could make it easier to use. Currently all these structs are decoupled.
The mentioned files could then also be moved to internal/.


internal/command/term/format.go
```diff
@@ -0,0 +1,18 @@
+package term
+
+import (
+	"fmt"
+	"time"
+)
+
+func BytesToMib(bytes uint64) string {
```
FranciscoKurpiel, line 8: Optional: Wouldn't it make sense to replace this function by

```go
func FormatSize(sizeInBytes int64) {
    if sizeInBytes < 1024 {
        return fmt.Sprintf("%d B", sizeInBytes)
    } else if sizeInBytes < 1024 * 1024 {
        return fmt.Sprintf("%.3f kiB", float64(bytes)/1024)
    } else if sizeInBytes < 1024 * 1024 * 1024 {
        return fmt.Sprintf("%.3f MiB", float64(bytes)/1024)
    } else ...
}
```

Maybe receive the `B` part is parameter, so it would also be usable for transfer rates (eg: `MiB/s`).

internal/command/term/format.go
```diff
@@ -0,0 +1,18 @@
...
+	return fmt.Sprintf("%.3f", float64(bytes)/1024/1024)
+}
+
+func DurationToStrSeconds(duration time.Duration) string {
```
FranciscoKurpiel, line 12: Optional: Same thing as the one above: make a `FormatTime` function that return `1min 2sec` or something like, instead of 62.

routines/pool.go
```diff
@@ -0,0 +1,138 @@
...
+	terminateWg sync.WaitGroup
+}
+
+type work struct {
+	fn       WorkFn
+	userData interface{}
+}
+
+// WorkFn is a function that is executed by the pool workers.
+type WorkFn func(userData interface{})
```
FranciscoKurpiel, line 26: I don't think we need the userData parameter. The caller can always provide data by using a _closure_ or an object, and those approaches would not degenerate the types to `interface{}`. If we do that then we don't even need the `work` struct anymore.

fho: That was the first version and then I stumbled about the issue that I was referencing in the closure a variable which changed outside of the closure and got very unexpected behavior, like:

```go
for _, build := range builds {
    pool.Queue(func() { fmt.Printf("build: %+v\n", build) })
}
```

build was already updated before the go routine run and all go-routines were running only for the last `build` in the `builds` slice

Do you maybe have a better idea to prevent the issue?

FranciscoKurpiel: I don't think accepting `interface{}` as parameter fixes the issue, because you can still pass references and change them, and you can still pass a closure, even when accepting the parameter. To me it looks like accepting the parameter makes the code more complicated and induces a weakening of the type system.

I think any approach will require the caller to make sure it don't modify anything being used on the job. We can have this without the need to add parameters.

fho: > I don't think accepting `interface{}` as parameter fixes the issue, because you can still pass references

It's for me non-intuitive that uploadAndRecord might be executed for `t.task` and `t.inputs` of other loop iterations in this example:
```go
for _, t := range taskToRun {
[...]
		c.uploadRoutinePool.Queue(
			func() {
				c.uploadAndRecord(ctx, t.task, t.inputs, outputs, runResult)
			})
	}
```

It's not intuitive for me because I don't pass a reference to `t` here.
It seems very easy to miss and error prone.

> and you can still pass a closure, even when accepting the parameter. 

true

> I think any approach will require the caller to make sure it don't modify anything being used on the job. We can have this without the need to add parameters.

Here is the `internal/command/run.go` code with a pool version that only accepts a function as parameter:
https://github.com/simplesurance/baur/blob/0de2dd2941707418eb81febb0aa9cecced5aacb8/internal/command/run.go#L274

Do you have another suggestion how I can pass in `t` here to prevent the described issue?

The only idea I have is to wrap the `Queue()` call  in another function to create a copy of `t`, like:
```go
func(t *pendingTask) {                                                                                                                                                                                                                                                   
       c.uploadRoutinePool.Queue(func() {                                                                                                                                                                                                     
              c.uploadAndRecord(ctx, t.task, t.inputs, outputs, runResult)                                                                                                         
       })                                                                                                                                                             
 }(t)
```

FranciscoKurpiel: I like you second approach. Another way would be to make sure the closure captures a variable that is declared inside the loop, effectively given each closure it's own variable with:

```go
for _, t := range taskToRun {
    t := t // make sure each each closure captures it's own variable
    [...]
}
```

See
https://play.golang.org/p/9Y-VBxT6IwR

fho: Thanks, I changed it as suggested.
You convinced me that the userData parameter did not made it better only more complicated :-)

internal/command/term/format.go
fho: brilliant :-)
thanks!

internal/command/term/format.go
fho: done, thanks!

scorsi:
Looks promising, hope to dive into it ! 🚀

PR #163, opened 2020-06-30
fix: recording run fails if baur-repository is not a git repository

fho:
run: fix: recording run fails if baur-repository is not a git repository
        
        When the baur repository was not a git repository, recording a task run failed
        because the HEAD git commit could not be resolved.
        
        How the repository state in the commands is refactored.
        The git package is moved to vcs/git.
        It got a new function that returns if a directory is part of a git repository.
        
        A vcs package is introduced with a function vcs.GetState(path).
        If path is in a git repository, it returns a git.RepositoryState object.
        Otherwise a dummy RepositoryState() object, implementing the same interface,
        that always returns vcs.ErrVCSRepositoryNotExist
        vcs.GetState() caches the returned RepositoryState() for the dir that it was
        called it.
        The vcs package will also make it easier to add support for other VCS in the
        future.
        
        If vcs.ErrVCSRepositoryNotExist is returned by those methods when returning a
        task run it is ignored and no vcs information are recorded.


fho:
Back to draft mode, had yesterday the idea that it should only instantiate git.RepositoryState if the `git` tools are in path.
It would also be great to log a debug message about the result (git repository found/not found, tools not installed, etc)

internal/command/helpers.go
```diff
@@ -124,8 +124,15 @@ func mustNewCompatibleStorage(r *baur.Repository) storage.Storer {
 	return clt
 }
 
+func mustGetRepoState(dir string) vcs.StateFetcher {
+	s, err := vcs.GetState(dir, log.Debugf)
+	exitOnErr(err, "evaluating if baur repository is in a VCS repository failed")
```
FranciscoKurpiel, line 129: I suggest rephrasing this.

fho: fixed

fho:
Created an issue to decide and move more packages to `internal/` https://github.com/simplesurance/baur/issues/165

fho:
merged

PR #166, opened 2020-07-09
cfg: comment every line in example app and include files

fho:
cfg: comment every line in example app and include files
        
        "baur init app" and "baur init include" generated example configuration files
        that had a minimal set of fields uncommented. baur was able to parse and use the
        generated example configs.
        E.g. a config file created with "baur init app", were showing up in `baur ls
        apps` after it's creation.
        
        Fields were commented by setting the `commented` for the fields in the struct.
        This made it impossible to initialize a config struct with some values, marshal it
        to disk, unmarshal it from disk again and get the same config struct.
        Because some fields were commented out in the file version, the unmarshaled
        version differs.
        
        To be able to marshal/unmarshal configs into equal objects:
        
        - uncomment all cfg struct fields,
        - add an option to the toFile function, to comment every line in the file,
        - comment all lines in the example app and include files created by "baur init
          app", "baur init include",
        
        This functionality is required to write integration testcases.


cfg/cfg.go
```diff
@@ -1,39 +1,97 @@
...
 
 	"github.com/pelletier/go-toml"
 )
 
+type toFileOpts struct {
+	overwrite bool
+	commented bool
+}
+
+// TToFileOpt is an option that can be passed to the ToFile functions
```
FranciscoKurpiel, line 18: Typo

cfg/cfg.go
```diff
@@ -1,39 +1,97 @@
...
-	encoder.Order(toml.OrderPreserve)
-	err = encoder.Encode(data)
+	f, err := os.OpenFile(filepath, fileOpenFlags(settings.overwrite), 0640)
 	if err != nil {
-		f.Close()
 		return err
 	}
 
+	if settings.commented {
+		if err := writeCommented(&buf, f); err != nil {
```
FranciscoKurpiel, line 58: One tiny optional thing: Use same parameter order as `io.Copy()`.

fho: thx, fixed

cfg/cfg.go
fho: fixed

fho:
merged manually

PR #168, opened 2020-07-13
Add first integration testcases

fho:
The PR introduces first integration testcases for the `baur run,` baur status`, `baur ls apps` commands.

The integration testcase are creating a baur repository in a temporary directory, create a new postgres database with a unique name per testcase and verify that the baur commands work as expected.

See the commit message for more detailed information.


fho:
@FranciscoKurpiel 

> I noticed you're not using the _test package name for tests. I think that's a great thing for limiting the risk of test dependencies being considered when doing non-test builds.

Then I would have to make things public in the command package only to be able to use it in tests.
Currently everything is private in the command package.

> For the future, it would be awesome to see testable examples, and having the example projects being used to test baur.

:+1: yes, would be great to also ensure in CI that basic commands in the example repo work as expected

fho:
merged manually

PR #181, opened 2020-09-01
Gosource input resolver: support to resolve testcase, new query syntax and specifying buildflags

fho:
The gosource input resolver is refactored:
- the sources of go test files can now be resolved,
- buildflags that are honored during resolving files, e.g. go buildtags can now be specified,
- instead of by path, go source are now specified via a query syntax

See the commit message for the full list of changes:
```
 init include: ensure filename has ".toml" suffix
        
        Check if the filename that is passed to "baur init include", has a ".toml"
        suffix. If it does not, add it.
        
        resolver: gosource: forward context
        
        Allow to pass the context through the resolver to packages.Load().
        This will allow to abort resolving when the context expires. This will be useful
        when setting timeouts for command operations.
        
        resolver: gosource: support to specify buildflags
        
        BuildFlags can now be specified when resolving golang sources.
        The specified BuildFlags are forwarded to packages.Load() which passes it as
        commandline parameters to the called build tool.
        
        This allows to e.g. specify build tags to only resolve sourcefiles which have
        "unittest" build tag.
        
        init app: allow creation of example .app.toml file outside of repository
        
        Do not check when running "baur init app" if we are in a baur-repository.
        It's not required to be within a baur repository to write an example .app.toml
        file. It's an unnecessary restriction.
        
        resolver: gosource: support fileglob queries
        
        Support a fileglob query type.
        When gosource resolver queries have the prefix "fileglob=", the glob pattern is
        resolved to files. The doublestar glob is supported.
        The resolved files are passed to the packages.Load() as filepaths.
        
        This allows to e.g. resolve the sources of all **/*_test.go files in
        go-unit-test tasks.
        
        resolver: gosource: move finding and checking if goroot exist to function
        
        resolver: gosource: do not set GOROOT env var, when getting it from "go env"
        
        When the gosource resolver got the GOROOT path from calling "go env", it's not
        needed to set the GOROOT environment variable to it.
        
        The go build tool will use the same path.
        
  fs: replace errors package with %w fmt specifier
        
        resolve: move resolving double-star globs to fs package
        
        Move the functions to resolve double star glob patterns from the resolve to the
        fs package.
        This will allow to reuse the functionality in other parts of the code.
        
        make: tests: disable verbose output
        
        Do not pass "-v" to "go test", the output became too much.
        
        resolver: gosource: support to query for go test files
        
        Add new configration field called Tests for the gosource resolver.
        If it's set to true, queries also resolve files to golang test files.
        By default this is disabled.
        
        This is useful when defining a baur task to run golang tests. With the option
        it's possible to include the test files as inputs.
        
        fs: DirsExist(): check for errors that are not of Type NotExist
        
        DirsExist() did not check if IsDir() returned an error that is not of type
        os.IsNotExist().
        This could cause some surprising results when IsDir() failed with another error.
        Do a proper error check.
        
        resolver: gosource: accept Queries instead of Paths
        
        The gosource resolver now accepts a list of queries instead of app-directory
        relative paths.
        This allows more freedom when specifying the go files for which sources should
        be resolved.
        
        The queries are passed through to packages.Load().
        It supports the same syntax then described at:
        https://github.com/golang/tools/blob/bc8aaaa29e0665201b38fa5cb5d47826788fa249/go/packages/doc.go#L17.
        
        - file=X to match files,
        - query=X and queries with a '<something>=' prefix are passed through directly
          to the underlying build tool (go list normally)
        
        Normally "go list" is used as build tool, which means the usual golang package
        patterns can be used.
        
        All queries are executed in the app directory.
        All defines Queries in the slice are passed to pacakges.Load() at once.
        Previously we called packages.Load() for each element in the Paths slice.
        
        The cfg files are adapted accordingly.
        When upgrading configs from v4, Paths are converted as:
        "./<OLD-PATH>/...". This should ensure the same behavior then before.
        
        resolver: gosource: log debug messages from packages package
        
        vendor: update golang.org/x/tools/go/packages
        
        resolver: gosource: replace errors package
        
        Use Golangs errors wrapping technique instead of the errors package


FranciscoKurpiel (approved):
Nice approach for the tests, but they are polluting the PRs quite a bit. There are about 100 testdata files, then `taskstatusevaluator.go`, then 100 vendor files. I can't come up with good approach that don't involve downloading test data (which is horrible) or submodules (wich I'm ok with, but some people find horrible).

PR #182, opened 2020-09-01
cfg: fix: app.Resolve() changes values of same included task instance

fho:
>         cfg: fix: app.Resolve() changes values of same included task instance
>         
>         When variables of an included task were resolved, values changed in the same
>         slice instance.
>         The values changed for independent App instances were affected.
>         
>         Fix it by deep copying the Task Include struct into a Task struct.
>         
>         The deep copy is done via gob encoding/decoding. It's comparatively slow but
>         works reliable and was fast to implement.
>         Copying the values in the struct would have much better performance but is more
>         error prone. When fields changes in the cfg it must not be forgotten to adapt
>         the copy logic.
>         This could be solved by a testcase but then it becomes much more effort then
>         using gob.
>         
> -------------------------------------------------------------------------------
>         add testcase to reproduce variable resolved to wrong val in include
>         
>         Add a testcase for the following issue:
>         
>         Scenario:
>         - 1 task includes include the same input include,
>         - the input-include uses a variable  like $APPNAME
>         
>         When 2 apps are run that include this task, the variable value is the $APPNAME
>         of the 1 app, when the task of the 2. app is executed.
>         
>


FranciscoKurpiel (approved):
Wouldn't it be better to use an approach based immutability? I mean, making `Input` and `Output` immutable, and making everything that changes them methods that return a changed copy, keeping the original in place? The approach this PR uses requires that the caller be aware that it should not modify objects, so it must clone. Since `Input` and `Output` are part of the public interface, we don't control the caller. Immutability seems to be necessary to me. Assuming we decide to use this approach, it would be totally fine to do it in future PR, because it would require changing more than a couple of lines.

The approach used here (deep-copy before change) looks good. I would prefer implementing a `Clone` method instead of using a deep copy library to require the structures being copied to explicitly specifying how they are copied, thus making them aware they are being copied, avoiding possible issues on the future regarding accidental copies of semaphores, times, os handles, counters and so on. But it is still totally fine to me.

fho:
> Wouldn't it be better to use an approach based immutability? I mean, making `Input` and `Output` immutable, and making everything that changes them methods that return a changed copy, keeping the original in place? 

Good idea, making the include files immutable would be sufficient.
the problem I see with making it immutable is that it's lot of overhead in code.
We would need to have accessor functions for the all fields.

We would save some copy operations and some memory but both are currently not relevant.
Copying the objects solves it with much less code and no noticeable disadvantage.


> Since Input and Output are part of the public interface, we don't control the caller. Immutability seems to be necessary to me. 

:+1: 
I'll look into it, making more methods private or returning copies from the IncludeDb could fix that.
Then the includes can be modified by the caller without unexpected side-effects

nocive (approved):

fho:
merged manually

PR #183, opened 2020-09-02
run: remove support for setting docker credentials via environment vars

### LINKED ISSUE #174: Resolver: Gosource: support to resolve only test files
Support to resolve only the input files of go-test files via the gosource-resolver.
This is useful to have a baur task to run go-tests which uses the gosource resolver to resolve inputs of the test files.

Current plan:
  - add a boolean tests config parameter to enable querying test files, it                                                                                                                                                                                     
    can be passed through to the packages resolver config                        
  - allow to specify queries instead of Paths for the gosource resolver,         
    pass it to packages.Load() directly                                          
  - either add a buildflags config parameter to enable setting buildflags or     
    check if it's possible to set them via env vars.                               
    It should allow to specify buildtags, for e.g. tag=integrationtest.            
  - add a query mode "baurglob=", that allows to specifies file globs, e.g.        
    */_test.go, to match all test files                                            
  - add tests (make use of testdata dir)

fho:
…riables

Setting docker credentials via the environment variables BAUR_DOCKER_USERNAME,
BAUR_DOCKER_PASSWORD is not supported it anymore.

Settings creds via environment variables is unsafe.
Credentials must be set instead via the "docker login" command.

Closes #174

### DIFF
--- a/internal/command/run.go
+++ b/internal/command/run.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"math"
-	"os"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


fho:
merged manually

PR #188, opened 2020-09-15
update baur0 dependency to version 0.18.1

fho:
This fixes issues when running baur on non-unix operation systems
related pr: https://github.com/simplesurance/baur/pull/186


fho:
fyi @StephenDiligent

PR #191, opened 2020-09-22
Improve input string handling

fho:
```
        command: rename --alt-input-str parameter to --lookup-input-str
        
        The name makes it more clear that the input str is only used for lookups.
        Fields and variable names are also adapted to the new name
        
        simplify handling input Strings
        
        Use the introduced Input interface to store Inputfiles and the inputString in
        the same slice.
        This simplifies the usage, conditional handling for the inputString can be
        removed in many places.
        
        InputString are now only added to Inputs if the string is not empty.
        Previously InputString with an empty String value were added which made it
        necessary to check later again if the String was empty or not.
        
        In some places inputs.GetInputString().Exists() was called without checking if
        GetInputString() returned nil.
        This could have caused a nil pointer dereference. This is fixed by this change.
        
        The commit also applies some minor cosmetic improvements.
```

Follow-up PR for: https://github.com/simplesurance/baur/pull/185

### DIFF
--- a/input.go
+++ b/input.go
@@ -5,4 +5,16 @@ import "github.com/simplesurance/baur/v1/internal/digest"
 // Input represents an input
 type Input interface {
 	Digest() (*digest.Digest, error)
+	String() string
+}
+
+// InputAddStrIfNotEmpty returns a new slice that contains in and str as an
+// InputString object, if str is not an empty string.
+// If it is, in is returned.
+func InputAddStrIfNotEmpty(in []Input, str string) []Input {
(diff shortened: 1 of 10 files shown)


fho:
@StephenDiligent fyi, feedback welcome :-)

StephenDiligent:
> @StephenDiligent fyi, feedback welcome :-)

I like it. It'll be much more extendable if further input types are to be added.

FranciscoKurpiel (approved):

Issue #180, opened 2020-09-01
resolver: add option to **not** fail if a resolver resolved to 0 files


fho:
Add a flag for the input resolves to not fail if a resolved found 0 files.

Currently if a resolver pattern matches 0 files, baur will fail with an error.

This makes it easier to manage inputs that only exist for some app.
With this flag, the inputs could be added to all apps, make the maintenance easier.

PR #192, opened 2020-09-28
cfg: convert Fileinput and GitFileInput fields to a slice

fho:
```
        cfg: convert FileInputs and GitFileInputs fields to a slice

        Converting them to a slice will allow to add additional fields to the
        struct that control how the input files are evaluated

        cfg: clone input/output includes before merging them with a config

        cfg: make all IncludeDB methods private

        The methods are only needed to be called internally, make them private.

```
The fields are converted to a slice to be able to specify different options per FileInput/GitFileInput section in the configs.
For https://github.com/simplesurance/baur/issues/180 new options will be added to those sections


FranciscoKurpiel (commented):
Please correct me if I'm wrong, but looks like this PR makes changes the configuration file syntax. But the change appears to me like something we want to do only internally. Or do we want to set custom attributes other than file path in the config file as well? And isn't this change a reason to have a new config file version (v5)?

fho:
> Please correct me if I'm wrong, but looks like this PR makes changes the configuration file syntax. But the change appears to me like something we want to do only internally. Or do we want to set custom attributes other than file path in the config file as well? 

Yes the attributes will be set in the configuration files.

> And isn't this change a reason to have a new config file version (v5)?

No, master is still unreleased and the config version nr in master  was already incremented compared to the latest release.

FranciscoKurpiel (approved):

PR #194, opened 2020-09-28
cfg: add Optional field to FileInputs and GitFileInputs sections

fho:
```
        cfg: add Optional field to FileInputs and GitFileInputs sections

        Add a new field called "optional" to the FileInputs and GitFileInputs cfg
        sections.
        By default it's value is false.
        If it's set to true, a Path that resolves to 0 files will not generate an error.
        make: add dbtest target

        The target runs all tests, including the one requiring a postgresql server

        ci: run tests with race detector enabled

        test: add function to initialize a new git repository

```


FranciscoKurpiel (approved):

fho:
merged manually

PR #195, opened 2020-09-29
cfg: rename S3Upload.Destfile field to S3Upload.Key

fho:
Use the naming schema from the s3 documentation instead of coming up with our
own.
https://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html#BasicsKeys

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,4 +20,4 @@ linters:
 
 linters-settings:
   goimports:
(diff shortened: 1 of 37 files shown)


FranciscoKurpiel (approved):
Change looks good, assuming the file conflict is minor.

fho:
merged manually

PR #197, opened 2020-09-29
Remove the InputsUndefined run status

### LINKED ISSUE #59: Refactor BuildStatusInputsUndefined status
When an application has no build inputs defined, baur assigns the status `BuildStatusInputsUndefined`.

Replace the status with a status that indicates that the build config for the application is incomplete.
The status should be assigned when the application either has:
- no inputs defined,
- no outputs defined or
- has outputs defined but no upload methods.

This will allow to not build applications via the `baur build` commands that are not recorded to the database by default, by only building application in the pending state

fho:
```
    remove the InputsUndefined status

        If a task has no inputs defined it is now treated as validation error.
        There is no benefit of handling this case special and have InputsUndefined
        status for it.

        Because having no inputs defined is now a validation error, the app configs must
        be merged with their includes before the validation is done.
        Before it was possible to do the validation before the merge without getting
        false errors.

        Validating the upgraded configuration files is moved to the command
        implementation. The validation is the normal process of loading all apps in the
        repository.

        upgrade: validate old cfg file before upgrading it

        cfg: remove redundant type definitions in slice declarations

        cfg: validate that the name of a task is not empty

```

This closes #59

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,4 +20,4 @@ linters:
 
 linters-settings:
   goimports:
-    local-prefixes: github.com/simplesurance/baur
+    local-prefixes: github.com/simplesurance/baur/v1
--- a/app.go
+++ b/app.go
@@ -1,7 +1,6 @@
 package baur
 
 import (
-	"path"
 	"path/filepath"
 	"sort"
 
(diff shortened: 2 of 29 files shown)


FranciscoKurpiel (approved):
Nice cleanup!

PR #200, opened 2020-09-29
cfg: run task commands not via the sh shell

fho:
Commands specified for a baur task where specified as a single string and always run via the sh-shell.
This allowed to use shell features like globs, etc.

The execution is changed to run the command directly instead.
The command and it's argument are specified in an array, exec-style.

This has the following advantages:

- command execution works on operating systems that do not have a sh-shell, like
  windows,
- the user can choose if a command should be run via a shell or if it should not,
- no unnecessary minimal overhead when shell functionality is not needed

### DIFF
--- a/internal/command/show.go
+++ b/internal/command/show.go
@@ -108,7 +108,7 @@ func (*showCmd) strCmd(cmd []string) string {
 	var result strings.Builder
 
 	for i, e := range cmd {
-		result.WriteString(fmt.Sprintf("\"%s\"", e))
+		result.WriteString(fmt.Sprintf("'%s'", e))
 		if i < len(cmd)+1 {
 			result.WriteRune(' ')
 		}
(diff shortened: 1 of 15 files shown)


internal/command/show.go
```diff
@@ -89,6 +89,19 @@ func mustWriteStringSliceRows(fmt format.Formatter, header string, indentlvl int
 	}
 }
 
+func (*showCmd) strCmd(cmd []string) string {
+	var result strings.Builder
+
+	for i, e := range cmd {
+		result.WriteString(fmt.Sprintf("\"%s\"", e))
```
FranciscoKurpiel, line 96: Shouldn't we escape `"` on `e` here? Like:
```go
result.WriteString(fmt.Sprintf("\"%s\"", strings.Replace(e, `"`, `\"`)))
```

Also, it might make sense to include tests with space and the `"` chars.

FranciscoKurpiel (commented):
Nice change, this way of representing parameters is a lot better. It a shame that we loose the two tests.

FranciscoKurpiel (approved):

fho:
merged manually

PR #201, opened 2020-10-01
Fix wrong input string being recorded against run

StephenDiligent:
Updating the implementation to match the comment (and fix the problem)


fho (approved):

fho:
thanks

PR #208, opened 2020-10-08
digest: fix: leading zero was truncating from sha384 string representation

fho:
@StephenDiligent figured out, that creating a sha384 digest and converting
it's string representation back to a digest can fail with a "hash length is 95,
expected length 96" error.

If a calculated digest started with 0, the leading 0 was truncated from the
hash.
This happened because the hash was stored as big.Int, leading zeroes can be
omitted for an integer without changing the meaning. Truncating a zero from a
digest makes it invalid.

Fix this by using the helpers from the hex package to create a string
representation of the hash instead of big.Int.

A testcase is added for the bug.
The existing testcase are changed to compare digests via their string
representation instead of converting the Sum to a string on their own.

### DIFF
--- a/internal/digest/digest.go
+++ b/internal/digest/digest.go
@@ -1,8 +1,8 @@
 package digest
 
 import (
+	"encoding/hex"
 	"fmt"
-	"math/big"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


FranciscoKurpiel (approved):
Looks good! :100:

Nice catch, @StephenDiligent!

PR #209, opened 2020-10-09
Handle digests with incorrect length that have already been stored in DB

StephenDiligent:
@fho Your changes in simplesurance/baur#208 will stop any new digests from being created with a length of 95 but there may already be digests that have been recorded against existing runs in the DB that have a length of 95.
This PR will account for these digests and append the leading zero that is missing.

### DIFF
--- a/internal/digest/digest.go
+++ b/internal/digest/digest.go
@@ -60,7 +60,12 @@ func FromString(in string) (*Digest, error) {
 
 		algorithm = SHA256
 	case "sha384":
-		if len(spl[1]) != 96 {
+		// There was a bug that caused digests that are created with a leading zero to be stored
+		// in the DB without the leading zero. This bug has been fixed but these digests need to be
+		// accounted for by appending the missing zero back onto the start of the digest.
+		if len(spl[1]) == 95 {
+			spl[1] = fmt.Sprintf("0%s", spl[1])
+		} else if len(spl[1]) != 96 {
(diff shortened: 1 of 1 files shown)


fho:
It could have happened that a calculated digest had multiple leading zeroes. The bug would have caused that all leading zeroes would be truncated. In this case we would need to prepend multiple zeros. 

I also think that is the wrong place for handling digests that were calculated wrongly in the past.
If a string that should be converted to a digest has <96 characters because of another reason or bug we would also prepend zeroes and end up with a wrong digest.

The bug can probably only cause issues when reading digests from the db and then converting it from string back to the Digest object. The right way to handle it would be to add missing leading zeroes to all stored digests in the db that are less then 95 chars long.
This would be a onetime operation and the length check in digest.go could be kept like it is.

The version in the master branch is the unstable development version. We do not need to provide workarounds for bugs in past unstable version.

We should fix it in the 0.18.x version though.  :-)
I suggest to backport the fix for 0.18, change the error message to contain a hint that it could have been triggered by this bug and a link to the release notes.
In the release notes we describe how the wrongly stored digests can be fixed in the database manually via an SQL statement.

Having a command to fix the issue in the db automatically would be even better, but I don't think it's worth the effort.
The likeliness to run into it seem low, nobody complained about it so far and we also did not run into it despite doing a lot of baur builds everyday.

fho:
@StephenDiligent I created a branch for version 0.18.2: https://github.com/simplesurance/baur/tree/v0.18.2

PR #215, opened 2020-10-19
Port to Windows

StephenDiligent:
This change set ports baur to Windows. The `make release` command will create a _baur.exe_ file.
Things to note:
1. A [breaking change](https://github.com/golang/go/issues/34610) was introduced to golang.org/x/sys that prevents github.com/docker/docker from compiling. This issue in github.com/docker/docker will not be [fixed](https://github.com/moby/moby/pull/40021) until v20 of github.com/docker/docker is released. In the meantime I have pinned golang.org/x/sys to the latest version before the breaking change was introduced.
2. Some of the existing tests used commands specific to Unix. These have either been replaced with equivalent commands available directly on both Unix and Windows or run via `cmd /C` on Windows.
3. Windows does not allow files that are in use to be deleted in the same way as Unix. In some tests files and folders had to be released so the test cleanup succeeds.
4. Expected paths have been updated to use OS specific path separators where required.
5. CircleCI does not provide Windows containers but does provide Windows VMs
    * The pre-installed version of golang is 1.12.7, this needs to be updated on every build. See [software-pre-installed-in-the-windows-image](https://circleci.com/docs/2.0/hello-world-windows/#software-pre-installed-in-the-windows-image)
    * I could not access Unix docker containers from within the Windows job so postgres is installed on the VM via chocolatey and the default "postgres" database is used.
    * The tests are much slower to run on Windows than Unix, the test timeout has been increased to accommodate this.
    * The `-race` flag has been removed from the Windows `go test` command as a gcc compiler is not installed. I figured this wasn't a problem as it is used in the existing test job. This could be addressed later if necessary.


StephenDiligent:
@fho Do I need to open a new issue for this work?

fho:
@StephenDiligent no, the PR is sufficient

fho:
Howdy @StephenDiligent :-)

I wondered why the Circleci builds were also not starting for this PR  and asked the CircleCI support:

> CircleCI prevents redundant builds from rerunning, which is why there is no build created for that PR. The PR submitter has their own project building on CircleCI, therefore when the PR is created, it does not trigger another build.
This causes issues with projects that require status checks to be complete. For now, they should stop building their project on their personal CircleCI account so that the build will run using the "Build forked pull requests"
option. When they create a new pull request, a CircleCI build will be triggered and they will then be able to see if they have passed their tests or not. We are investigating these types of situations as we understand this is not
the best-case scenario.

Could you try to disable in your CircleCI builds for your fork and create a new PR?

StephenDiligent:
> Howdy @StephenDiligent :-)
> 
> I wondered why the Circleci builds were also not starting for this PR and asked the CircleCI support:
> 
> > CircleCI prevents redundant builds from rerunning, which is why there is no build created for that PR. The PR submitter has their own project building on CircleCI, therefore when the PR is created, it does not trigger another build.
> > This causes issues with projects that require status checks to be complete. For now, they should stop building their project on their personal CircleCI account so that the build will run using the "Build forked pull requests"
> > option. When they create a new pull request, a CircleCI build will be triggered and they will then be able to see if they have passed their tests or not. We are investigating these types of situations as we understand this is not
> > the best-case scenario.
> 
> Could you try to disable in your CircleCI builds for your fork and create a new PR?

Sure thing, I'll do it now.

PR #218, opened 2020-10-27
Tests: log all output of commands via the test logger

fho:
This PR redirects output that is generated during test execution to the log functions of `testing.T`.
Forwarding the output to the test logger allows `go test` to associate output with the associated testcases.
This makes it easier to debug issues when tests are failing.

So far a lot of the output generated during test execution was printed simply to stdout or stderr and `go test` showed it intermingled with output from other tests.

1. The log package is changed to support forwarding all output to the test logger
2. A wrapper io.Writer wrapper for the testlogger is introduced that can be used as stdin and stdout streams for the commands.
3. Testcases are adapted to change the logger and stdin/stdout streams to log via the test logger.

### DIFF
--- a/cfg/app.go
+++ b/cfg/app.go
@@ -102,7 +102,7 @@ func (a *App) ToFile(filepath string, opts ...ToFileOpt) error {
 	return toFile(a, filepath, opts...)
 }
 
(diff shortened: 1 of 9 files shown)


FranciscoKurpiel (approved):
Looks great!

PR #226, opened 2020-11-12
docker: improve choosing credentials for registry, support URLs

### LINKED ISSUE #225: cherry-pick https://github.com/simplesurance/baur/pull/224 to baur/master
cherry-pick https://github.com/simplesurance/baur/pull/224 to baur/master

fho:
**This is a cherry-pick of https://github.com/simplesurance/baur/pull/224 for master**

Closes #225 
-------

"docker login" supports to specify registry per hostname and as URL.
If the credentials used the URL, baur was not able to find the credentials when
the push was done using only the hostname of the registry server.
This is a common usecase for GCR auth per json_key.

The authentication logic was also approved in the following ways:
- registry URLs with ports are supported,
  If a registry was authenticated with the default port 5000), and no port is
  given for the registry url in the app config, the credentials for the
  registry with the default port is used
- the credentials for the default registry (index.docker.io) are only used when
  the host of the specified registry matches or no registry is given in the app
  config
- when no credentials were found for registry, none are used instead of the
  first found one, this was a security issue, credentials might have been send
  to a wrong server

### DIFF
--- a/internal/upload/docker/docker.go
+++ b/internal/upload/docker/docker.go
@@ -5,12 +5,14 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"net/url"
 
 	docker "github.com/fsouza/go-dockerclient"
 )
 
 // DefaultRegistry is the registry for that authentication data is used
 const DefaultRegistry = "https://index.docker.io/v1/"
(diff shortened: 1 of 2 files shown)


FranciscoKurpiel (approved):
Great tests! :100:

PR #228, opened 2020-11-13
Gosourceresolver: Do not resolve to system-local input file paths for non-vendored go deps

### LINKED ISSUE #179: GolangSource resolver: make it work better with non-vendored dependencies
If a go.mod project uses non vendored dependencies, the input files are resolved to files in the `GOMODCACHE` dir. 
This directory is usually in `$HOME`.
Input files must only be resolved to files that are part of the repository.

Improve the behavior, ideas:
- get from `go list` the list of imported modules + their hash sum and use this string as input 
- remove $PKGDIR from paths that contain it, to have machine-independent paths

fho:
gosourceresolver: use $GOMODCACHE in paths for non-vendored external sources

        When a go.mod project did not vendored external libraries, baur resolved the
        source files of the libraries to the user's $GOMODCACHE directory.

        The $GOMODCACHE is normally in the home directory of the current user and
        outside of the baur repository.

        This means when the same application was build with another user or a different
        GOMODCACHE setting, the files resolved to different paths, baur rebuilded the
        applications unnecessarily.

        To make those paths independent of local paths, the gosourceresolver
        replaces paths that have start with the value of the $GOMODCACHE variable with
        "$GOMODCACHE".

Closes #179 

Related: https://github.com/golang/go/issues/42590


FranciscoKurpiel (approved):
Awesome tests! :1st_place_medal:

fho:
merged

PR #230, opened 2020-11-20
Cleanup the cfg package

fho:
- delete some unused methods
- make methods private if they do not need to be public


FranciscoKurpiel (approved):
:1st_place_medal:

Issue #232, opened 2020-12-06
baur diff: add input-str and lookup-input-str functionality to diff command


dbason:
The input str and diff commands are both looking really good, there is just one more piece to help round out our use case (https://github.com/simplesurance/baur/issues/117)

We would like to combine the two so that we can also use the input-str functionality when we are diffing the last build.  The two pieces of functionality would be:

- When input-str is added using the carrot(^) notation it will look up the builds with the input-str.  The diff will then use the input str to do the comparison
- When lookup-input-str is provided if no previous build is found with the input-str it will then do the search using the lookup-input-str.  For the purposes of comparison both the previous build and the current working directory will use the input-str.

Please let me know if anything needs further clarification

StephenDiligent:
@fho This issue has been addressed by the changes in [Add input-str and limit filtering to ls runs command](https://github.com/simplesurance/baur/pull/239 ).
We now have all of the required changes to address our use case. When can we expect the next release to be published?
FYI @dbrougham

fho:
@StephenDiligent hard to tell, depends on how many other more important things will come up at my work. :-)

All the major changes are implemented. There are only some smaller things left that I'll finish in the next days (https://github.com/simplesurance/baur/milestone/1).
I aim to have a first `-rc` release until end of the month. We will then already migrate our companies internal CI workflow to the new version and start to use it.

PR #233, opened 2020-12-06
Support input-str and lookup-input-str flags in diff inputs command

StephenDiligent:
This change set addresses https://github.com/simplesurance/baur/issues/232

I decided to add a new query to load task runs that contain an input with a given digest to find runs by their input-str instead of loading all task runs then iterating over them within baur. 

Example outputs from these changes:
A single run exists for baur_foo.build that contains an input-str of "master"
**run with given input-str does not exist**
image
**run with given input-str and lookup-input-str does not exist**
image
**run with given input-str does exist**
image
**run with given input-str does not exist but given lookup-input-str does exist**
image

@fho

### DIFF
--- a/internal/command/diff_inputs.go
+++ b/internal/command/diff_inputs.go
@@ -31,11 +31,15 @@ func (i *storageInput) String() string {
 }
 
 type diffInputArgDetails struct {
-	arg      string
-	appName  string
-	taskName string
-	runID    string
-	task     *baur.Task
+	arg                  string
+	appName              string
+	taskName             string
+	runID                string
+	inputStr             string
+	inputStrDigest       string
+	lookupInputStr       string
+	lookupInputStrDigest string
+	task                 *baur.Task
 }
 
 func init() {
@@ -45,9 +49,10 @@ func init() {
 type diffInputsCmd struct {
 	cobra.Command
 
-	csv      bool
-	quiet    bool
-	inputStr string
+	csv            bool
+	quiet          bool
+	inputStr       string
+	lookupInputStr string
 }
 
 const diffInputslongHelp = `
@@ -106,6 +111,9 @@ func newDiffInputsCmd() *diffInputsCmd {
 	cmd.Flags().StringVar(&cmd.inputStr, "input-str", "",
 		"include a string as input")
 
+	cmd.Flags().StringVar(&cmd.lookupInputStr, "lookup-input-str", "",
+		"if a previous run can not be found with the input-str provided, try to find a previous run with this value as input-string")
+
 	return &cmd
 }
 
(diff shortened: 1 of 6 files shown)


StephenDiligent:
It looks like one of the existing tests failed because it was unable to connect to the postgres DB. 
I do not have the necessary permissions to rerun the check in CircleCI.

fho:
@StephenDiligent
thanks for the PR, I'll have a look soon.

 I rerun the test and they are still failing with `FATAL: sorry, too many clients already`.
The issues should be that we invoke in the test the command implementation which do not close the psql connections currently.
With the new change the max. number of parallel psql connections was reached.
I'll add the close calls for the psql connection.

fho:
Sorry but I have problems to understand the change.
Could you please explain in the PR description the change, including what effect `--input-str` and `--lookup-input-str` has when specifying task, task-run ids, or task runs via the caret notation?

Please correct me if I'm wrong with my comments: :-)

The new behavior of the parameter seems inconsistent with how it works for other commands.
The usage description for `--input-str` is e.g. the same but it seems not to simply add the string as additional input for the first specified task-run/task, instead it changes the db-query?
In general the same parameter should have consistent behavior for all commands.
For example I would expect that:
`ngbaur diff inputs --input-str master random.build random.build^`
shows the difference of the inputs defined in the .app.toml file of the `random.build` task, with the additional string input `master`,  and the last recorded run of `random.build`.

It's also confusing to me that the positional parameters for `baur diff inputs` specify exactly which builds too compare.
When `--input-str` or `--lookup-input-str is passed though they to change their meaning and become queries . This behavior is  also not clearly described in the `baur diff inputs --help` output.

If you want to find the last recorded run of a task with a specific input string we might need to a new query syntax instead of changing the meaning.
Finding the last recorded task run with a specific input string might also fit well into `baur ls runs`, then the required functionality could be combined like:
```sh
run_id="$(baur ls runs --has-input-with-digest=sha256:123  --sort time-desc --csv --fields id baur_foo.build)
baur inputs diff baur_foo.build $run_id
```

dbason:
Hi Favian,

The idea to us `baur ls runs` and then pass that along sounds fine to me.

Our core use case is that we need to compare the current state on a PR or branch against the previous build for that PR or branch.

StephenDiligent:
I'm closing this PR in favour of adding `input-str` and `lookup-input-str` support to the `baur ls runs` command.

fho:
Hello @dbason @StephenDiligent,

I would suggest that you describe your usecase in a new post at the [Discussions Github page](https://github.com/simplesurance/baur/discussions?discussions_q=category%3AIdeas). If you already have ideas how it could be used from the commandline please, also add those as examples.
Then we figure out a way together how we could integrate what you need before starting the implementation.

Maybe we can also find a way to do what you need via "baur diff inputs" if that is a better workflow.

@dbason Yes, I agree regarding `--has-input-with-digest`. The user should not have to calculate the digest on his own.

dbason:
@fho I've added a discussion at https://github.com/simplesurance/baur/discussions/235

I hope that describes our process and use case in full.

Issue #236, opened 2020-12-14
diff inputs - Does not accept an app name that contains a dash '-'


dbrougham:
Hi,

When using the `diff inputs` command with an application name that contains a dash the command will throw an error.

Command: `baur diff inputs application-one.build application-one.build^`

![image](https://user-images.githubusercontent.com/3860517/102032431-9a61b380-3e1d-11eb-9ce1-a41aac490d7e.png)

If I rename the inputs to `application_one.build` the command will resolve correctly.

Command: `baur diff inputs application_one.build application_one.build^`

![image](https://user-images.githubusercontent.com/3860517/102032685-5de28780-3e1e-11eb-9b9f-d48f963d760f.png)

Have narrowed down the root cause to the regex expression used here: https://github.com/simplesurance/baur/blob/ba560e8f282c67f673ff865ce58a7f46e8a52ada/internal/command/diff_inputs.go#L112

My question is what additional characters should also be supported? Or should it allow any characters followed by a single period and then the task name?

fho:
Hello David,

thanks for the report. 
That is a bug, dashes should be supported supported in app and task names

> My question is what additional characters should also be supported? Or should it allow any characters followed by a single period and then the task name?

I would suggest `[\w-]+` for app and task names

PR #237, opened 2020-12-15
236-ALLOW-DASH-DIFF-INPUTS - Update diff inputs regex 

dbrougham:
Update regex used to allow application names that contain a dash character.

Fix for https://github.com/simplesurance/baur/issues/236


fho:
@dbrougham thanks for the fix.
I spotted another issue with the parameter validation during the review: https://github.com/simplesurance/baur/pull/238 :-)

dbrougham:
@fho thanks for the quick review and merge. Also nice catch there! 🙌

PR #239, opened 2021-01-05
Add input-str and limit filtering to ls runs command

StephenDiligent:
These changes implement the suggestions made in the [Find most recent build with a specifc input-str](https://github.com/simplesurance/baur/discussions/235) discussion.
A couple of things to note:

- To be able to support filtering by an input URI value I needed to introduce a new query that included the `input.uri` field in the query's inner `SELECT DISTINCT ON` clause. Without adding the `input.uri` field only a single input would be returned by the inner SELECT statement and it may not have been the input we wanted to filter on. If the `input.uri` field was added to the existing query any queries that did not filter by input URI would return a row for each input which is not the desired behaviour. 
- I opted to add a `has-input-str` flag instead of the `has-input` flag mentioned in the discussion. The reason for this is that the user would have to have internal knowledge of baur to know that a "string:" prefix would be added to the input-str values when they are stored in the DB to be able to filter by input-str. The `has-input` flag could still be added later to be able to filter by input file paths but it is not required for the use case in the discussion.

@fho

### DIFF
--- a/internal/command/ls_runs.go
+++ b/internal/command/ls_runs.go
@@ -240,7 +240,7 @@ func (c *lsRunsCmd) getFilters() []*storage.Filter {
 
 	if c.inputURI != "" {
 		filters = append(filters, &storage.Filter{
-			Field:    storage.FieldURI,
+			Field:    storage.FieldInput,
 			Operator: storage.OpEQ,
 			Value:    c.inputURI,
 		})
--- a/internal/command/ls_runs.go
+++ b/internal/command/ls_runs.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/simplesurance/baur/v1"
 	"github.com/simplesurance/baur/v1/internal/command/flag"
 	"github.com/simplesurance/baur/v1/internal/command/term"
 	"github.com/simplesurance/baur/v1/internal/format"
@@ -37,11 +38,12 @@ func init() {
 type lsRunsCmd struct {
 	cobra.Command
 
-	csv    bool
-	after  flag.DateTimeFlagValue
-	before flag.DateTimeFlagValue
-	sort   *flag.Sort
-	quiet  bool
+	csv      bool
+	after    flag.DateTimeFlagValue
+	before   flag.DateTimeFlagValue
+	inputStr string
+	sort     *flag.Sort
+	quiet    bool
 
 	app  string
 	task string
@@ -79,6 +81,9 @@ func newLsRunsCmd() *lsRunsCmd {
 
 	cmd.Flags().VarP(&cmd.before, "before", "b",
 		fmt.Sprintf("Only show runs that were started before this datetime.\nFormat: %s", term.Highlight(flag.DateTimeFormatDescr)))
+
+	cmd.Flags().StringVar(&cmd.inputStr, "has-input-str", "",
+		"Only show runs that include this value as an input-str value")
 
 	return &cmd
 }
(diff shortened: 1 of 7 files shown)


internal/command/ls_runs.go
```diff
@@ -37,11 +40,13 @@ func init() {
...
-	after  flag.DateTimeFlagValue
-	before flag.DateTimeFlagValue
-	sort   *flag.Sort
-	quiet  bool
+	csv      bool
+	after    flag.DateTimeFlagValue
+	before   flag.DateTimeFlagValue
+	inputStr string
+	sort     *flag.Sort
+	limit    int
```
fho, line 48:
```suggestion
	limit    uint
```

This would ensure that you can not pass invalid negative values on the command line for `--limit`

storage/storage.go
```diff
@@ -92,13 +96,30 @@ type Storer interface {
...
+	) error
+
+	// TaskRunsWithInputURI queries the storage for runs that match the filters
+	// and contain an input with the given URI value.
+	// A limit value of 0 will return all results.
+	// The found results are passed in iterative manner to the callback
+	// function. When the callback function returns an error, the iteration
+	// stops.
+	// When no matching records exist, the method returns ErrNotExist.
+	TaskRunsWithInputURI(ctx context.Context,
```
fho, line 118: That the postgres storage  implementation requires a different query when a filter for the `FieldURI` field is specified, is specify to the postgres implementation.
This complexity and detail should be hidden in the storage interface.

The `TaskRuns()`  function could check if a filter for the `FieldURI` was passed, if yes it can use a different query.
What do you think, @StephenDiligent?

storage/postgres/query.go
```diff
@@ -261,7 +316,6 @@ func (c *Client) TaskRuns(
 			&taskRun.StartTimestamp,
 			&taskRun.StopTimestamp,
 			&taskRun.Result,
-			nil, // skip scanning of duration value, it's only used for filtering and sorting
```
fho, line 264: :medal_military:   thx :-)

internal/command/ls_runs.go
```diff
@@ -28,7 +29,9 @@ baur ls runs -s duration-desc calc               list task runs of the calc
 						 application, sorted by
 						 run duration
 baur ls runs --csv --after=2018.09.27-11:30 '*'  list all task runs in csv format that
-						 were started after 2018.09.27 11:30`
+						 were started after 2018.09.27 11:30
+baur ls runs --limit=1 calc                      list a single task run of the calc
```
fho, line 33: indentation is off here

fho (commented):
@StephenDiligent thanks for the PR, looks great.

I added some suggestions for improvements.

> I opted to add a has-input-str flag instead of the has-input flag mentioned in the discussion. The reason for this is that the user would have to have internal knowledge of baur to know that a "string:" prefix would be added to the input-str values when they are stored in the DB to be able to filter by input-str. 

I'm fine with keeping the `--has-input-str` parameter if it makes the usage easier.
I'm not convinced that  `string:` is or should be internal knowledge though. :-)
`baur ls inputs` is also showing string inputs with the `string:` prefix currently.
Therefore having only a `--has-input` parameter and documenting in the usage output of the command that the `string:` prefix is supported would be ok.


> The has-input flag could still be added later to be able to filter by input file paths but it is not required for the use case in the discussion.

:+1: I can also add it in a follow-up PR. It should be simple after your groundwork. :-)

storage/storage.go
StephenDiligent: I think this is a great idea, I'l make this suggested change

internal/command/ls_runs.go
StephenDiligent: It appears that spaces instead of tabs have been used on the lines that begin with `baur ls`, I'll fix this up for all of the examples in this file.

StephenDiligent:
@fho Thank you for the review. I have addressed your comments and also refactored the `--has-input-str` flag into a `--has-input` flag.

storage/postgres/query.go
```diff
@@ -219,21 +221,46 @@ func (c *Client) TaskRuns(
...
+			"{distinct_on}", "task_run.id, input.uri",
+			"{fields}", "input.uri AS uri,",
+			"{joins}", "JOIN input ON task_run_input.input_id = input.id")
+	} else {
+		replacer = strings.NewReplacer(
+			"{distinct_on}", "task_run.id",
+			"{fields}", "",
+			"{joins}", "")
+	}
+	queryStr := replacer.Replace(queryTemplate)
```
fho, line 255: cool idea to prevent that we have to duplicate query strings :tada:  :-)

fho (approved):

fho:
@StephenDiligent thanks, it's merged! :-)

PR #240, opened 2021-01-05
Fix tests on Windows

StephenDiligent:
I noticed that some of the tests have been failing locally but passing on CircleCI. It turns out the `go test` command exit code was not being reported back to CircleCI correctly so the job was passing when it shouldn't have been.

I expect the checks for 5720734 to fail. I will then push up the fixes to the tests that are failing on Windows and the checks should then legitimately pass.

@fho


StephenDiligent:
I closed this PR because the checks were cancelled before the tests ran and reported back the required exit code to cause a failure.

PR #241, opened 2021-01-05
Fix tests failing on Windows

StephenDiligent:
I noticed that some of the tests have been failing locally but passing on CircleCI. It turns out the go test command exit code was not being reported back to CircleCI correctly so the job was passing when it shouldn't have been.
image

I expect the checks for 5720734 to fail. I will then push up the fixes to the tests that are failing on Windows and the checks should then legitimately pass.

@fho

### DIFF
--- a/internal/resolve/gosource/gosource_test.go
+++ b/internal/resolve/gosource/gosource_test.go
@@ -3,7 +3,6 @@ package gosource
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -15,15 +14,6 @@ import (
 	"github.com/simplesurance/baur/v1/internal/prettyprint"
 	"github.com/simplesurance/baur/v1/internal/testutils/strtest"
 )
-
-func containsGoPath(envs []string) bool {
-	for _, env := range envs {
-		if strings.Contains(env, "GOPATH") {
-			return true
-		}
-	}
-	return false
-}
 
 func TestResolve(t *testing.T) {
 	const testCfgFilename = "test_config.json"
@@ -59,14 +49,6 @@ func TestResolve(t *testing.T) {
 
 			cwd, err := os.Getwd()
 			require.NoError(t, err)
-
-			if os.Getenv("GOCACHE") == "" {
-				os.Setenv("GOCACHE", t.TempDir())
-			}
-
-			if !containsGoPath(testCfg.Cfg.Environment) {
-				testCfg.Cfg.Environment = append(testCfg.Cfg.Environment, fmt.Sprintf("GOPATH=%s", os.Getenv("GOPATH")))
-			}
 
 			for i := range testCfg.Cfg.Environment {
 				testCfg.Cfg.Environment[i] = strings.Replace(testCfg.Cfg.Environment[i], "$WORKDIR", cwd, -1)
--- a/internal/resolve/gosource/gosource_test.go
+++ b/internal/resolve/gosource/gosource_test.go
@@ -3,6 +3,7 @@ package gosource
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


internal/resolve/gosource/gosource_test.go
```diff
@@ -50,6 +60,14 @@ func TestResolve(t *testing.T) {
 			cwd, err := os.Getwd()
 			require.NoError(t, err)
 
+			if os.Getenv("GOCACHE") == "" {
+				os.Setenv("GOCACHE", t.TempDir())
+			}
+
+			if !containsGoPath(testCfg.Cfg.Environment) {
+				testCfg.Cfg.Environment = append(testCfg.Cfg.Environment, fmt.Sprintf("GOPATH=%s", os.Getenv("GOPATH")))
+			}
```
fho, line 69: Why is the `GOPATH` environment variable set if it's unset?
Shouldn't either be go modules be used, then GOPATH is ignored or go modules are not used and GOPATH should be set in the `test_config.json files`

internal/resolve/gosource/gosource_test.go
```diff
@@ -50,6 +60,14 @@ func TestResolve(t *testing.T) {
 			cwd, err := os.Getwd()
 			require.NoError(t, err)
 
+			if os.Getenv("GOCACHE") == "" {
+				os.Setenv("GOCACHE", t.TempDir())
```
fho, line 64: Why is GOCACHE set in the testcase?

StephenDiligent: Without specifying a GOCACHE value this error occurs `exec: running 'C:\Go\bin\go.exe env -json' in directory '.' exited with code 1, expected 0, output: 'build cache is required, but could not be located: GOCACHE is not defined and %LocalAppData% is not defined'`

internal/resolve/gosource/gosource_test.go
StephenDiligent: Without specifying a GOPATH value this error occurs `exec: running 'C:\Go\bin\go.exe env -json' in directory '.' exited with code 1, expected 0, output: 'missing $GOPATH'`
The reason it only sets a value if it's unset is because some tests already specify a GOPATH value.

internal/resolve/gosource/gosource_test.go
fho: Sounds to me like the root is that %LocalAppData% is not set in our execution environment and if we set it Go will be able to use the default GOCACHE directory.

fho: We probably have to whitelist `LocalAppData` here: 
https://github.com/simplesurance/baur/blob/d5dfe9cae22537dc6bd8ed592e5e996725ee6ac7/internal/resolve/gosource/gosource.go#L132

I'm  currently testing the change in a separate PR and push the commit to this branch if it works.

internal/resolve/gosource/gosource_test.go
fho: >  The reason it only sets a value if it's unset is because some tests already specify a GOPATH value.

But what is the reason that GOPATH must be set at all?
If the GOPATH environment variable is empty go determines a reasonable default on it's own normally.
Example on Linux:
```sh
$ unset GOPATH; go env GOPATH
/home/fho/go
```

fho: It failed because the `USERPROFILE` environment variable was not made available when running (`go env`).
(https://github.com/golang/go/wiki/SettingGOPATH).
I fixed it in https://github.com/simplesurance/baur/pull/241/commits/c6d2f22d683dd3c2871795c38a3270366db0303b and reverted changing the testcases

internal/resolve/gosource/gosource_test.go
fho: works :-)
https://github.com/simplesurance/baur/pull/241/commits/c6d2f22d683dd3c2871795c38a3270366db0303b

fho:
I merged the changes manually to master.

StephenDiligent:
@fho Thanks for sorting out the gosource issues in a Windows specific way!

PR #252, opened 2021-01-15
 cfg: refactor Input.GitFiles section

fho:
```
        cfg: refactor Input.GitFiles section

        The Input.GitFiles section in the config is removed.
        Instead a new option field is added to the Input.Files  section called
        git_tracked_only.
        When it is set to true, the resolver will only resolve to files that exist in
        the filesystem and are tracked by git.

        This makes the configuration file more simple, nesting is reduced and it also
        makes it more clear that the syntax for the paths is now the same for both.

        The commit also fixes a bug that when a glob like "*" was passed to the
        gitpathresolver and not all of the filepaths to that the glob resolved were
        tracked by git, an error was returned.
        This changed accidentally the previous behavior and is not wanted.
        The resolver should return only files that are in the filesystem **and** tracked
        by Git.
        This is fixed by running the resolver per entry in the Inputs.GitFiles.Path
        field in the app config and not passing "--error-unmatch" to git-ls files.
        The interface of the resolver is made more similar to the FilesResolver.
        It now accepts only a single path instead of a list.
        Checking if an error returned because a glob resolved to 0 files is also done in
        the caller instead of in the resolver.

        cfg: replace "." Input.GitFile paths with "**"

        Because GitFile input path patterns are now resolved via fs.FileGlob()
        instead of via git, the path "." resolved to 0 files.
        Previously it resolved to all files in the current and all subdirectories.

        Change "." paths to "**" when configs are upgraded. "**" behaves the same then
        then "." previously.

        git: return slice of files instead of output of the command

        git.LsFiles now splits output of ls-files into a slice of filepaths insteadf of
        returning the output of the command  as string.
        Previously the split was done in the caller, which did not make sense.

        cmd: remove duplicate information in error message

        When inputResolver.Resolve() fails the error message already contains a
        context like "resolving FileInputs" failed, do not add the additional message
        "resolving inputs failed" additionally to the error.

```


cfg/inputinclude.go
```diff
@@ -49,6 +44,7 @@ func (in *InputInclude) clone() *InputInclude {
 	var clone InputInclude
 
 	deepcopy.MustCopy(in, &clone)
+	// TODO why is filepath assignment manually and not cloned?
```
FranciscoKurpiel, line 47: Strings are immutable, so there is no need to clone.

FranciscoKurpiel (approved):
Looks great! :100:

cfg/inputinclude.go
fho: thanks :-)

PR #257, opened 2021-01-22
gosourceresolver: do not return files that are in the GOCACHE directory

### LINKED ISSUE #254: gosource resolver: resolves to go cache files when tests option is enabled
Example:

Running the gotestresolver for test files, resolved to: 
```
../../../.cache/go-build/17/17ad5263d0bb12e623f7deb7f763dacce722ded07d0a0662cf71813ad39efc1a-d
../../../.cache/go-build/1f/1f9522c2cd8211727a103193c5b10c646890a730e05d34de83d1a88893f96a7f-d
../../../.cache/go-build/24/245d116ce1e0d527c808311392fc28f38ccd7d48b57f64a545d1b1c1fbbb9635-d
../../../.cache/go-build/3e/3e148b7a06e493e20c46c4cf21de70eb54cfee7f6a478814e32d3370545a2c7d-d
../../../.cache/go-build/58/58a6f8e48452015886b09e036f868057601df99ce4e3ff89732c72479aefc802-d
../../../.cache/go-build/f1/f1ceeaf655456fa6b2ff7119eef890a97f6ea353fe2b4702b1c4e618b8cea589-d
```

The paths were invalid.
The resovler should not resolve to local cache files.

fho:
When a go test already run before, the gosourceresolver was returning paths
from the local GOCACHE directory.

Do not return files that are in the GOCACHE directory. We only want to return
source code files. Files in GOCACHE depend on the source files and are
regenerated reproducible for the same source files. They do not need to be
tracked as inputs.

Closes #254

### DIFF
--- a/internal/resolve/gosource/gosource.go
+++ b/internal/resolve/gosource/gosource.go
@@ -237,6 +237,11 @@ func withoutStdblibPackages(result *[]string, env *goEnv, paths []string) error 
 		if err != nil {
 			return err
 		}
+
+		if strings.Contains(abs, env.GoCache) {
+			continue
+		}
+
 		// use HasPrefix() + Replace() to ensure we only replace the
 		// path if is the prefix
 		if strings.HasPrefix(abs, env.GoModCache) {
--- a/internal/resolve/gosource/gosource.go
+++ b/internal/resolve/gosource/gosource.go
@@ -238,7 +238,7 @@ func withoutStdblibPackages(result *[]string, env *goEnv, paths []string) error 
 			return err
 		}
 
-		if strings.Contains(abs, env.GoCache) {
+		if strings.HasPrefix(abs, env.GoCache) {
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


internal/resolve/gosource/gosource.go
```diff
@@ -237,6 +237,11 @@ func withoutStdblibPackages(result *[]string, env *goEnv, paths []string) error
 		if err != nil {
 			return err
 		}
+
+		if strings.Contains(abs, env.GoCache) {
```
FranciscoKurpiel, line 241: Isn't this condition too weak? If cache is `/c', for example, this code will ignore `/my/code`.

FranciscoKurpiel (changes requested):

internal/resolve/gosource/gosource.go
fho: yes, thanks
I should check for `strings.HasPrefix`

fho: fixed

FranciscoKurpiel (approved):
Assuming `GoCache` is always an absolute path that looks good. :100:

fho:
> Assuming `GoCache` is always an absolute path that looks good. 100

You found more bugs, fixed :-)
thanks

FranciscoKurpiel:
Double-approved!

fho:
merged manually

PR #261, opened 2021-01-29
move public packages to pkg/ directory

fho:
All go code that acts as baur's public API or is imported in the command
implementation is moved to the pkg/ directory.

It makes the root directory look cleaner, apart from that there are no
advantages.


fho:
I finally implemented your suggestion you made a while ago, @FranciscoKurpiel  (-:

FranciscoKurpiel:
Yay!!!

Quick question: you opted to move `storage` to `pkg/`, not to `internal/'. Is it intentional?

fho:
> Yay!!!
> 
> Quick question: you opted to move `storage` to `pkg/`, not to `internal/'. Is it intentional?

Yes, I want to be able to import the storage and cfg package in a future version.
I might want to import them to use the package for converting v1 cfg or storage to a new format.
The go module would then change to github.com/simplesurance/baur/v2` and I would vendor for  `github.com/simplesurance/baur/v1/pkg/cfg´.

Same then what I'm doing here with v0 cfg pkg: https://github.com/simplesurance/baur/blob/68a4fe876505a14985fa09397bb57d3fa74bc1f2/cfg/upgrade/v4/v4.go#L7

FranciscoKurpiel (approved):

PR #287, opened 2021-03-15
Add .appdir variable

kostyay:
This is a handy variable if you want to define a task in shared include and reference the step directory from it

### DIFF
--- a/pkg/baur/cfg_resolvers.go
+++ b/pkg/baur/cfg_resolvers.go
@@ -2,6 +2,7 @@ package baur
 
 import (
 	"errors"
+	"path/filepath"
 
 	"github.com/simplesurance/baur/v1/internal/vcs"
 	"github.com/simplesurance/baur/v1/pkg/cfg/resolver"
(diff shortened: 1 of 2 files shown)


fho:
@kostyay thanks for the PR.

Tasks are always executed in the application directory also if they are defined in an include file.
Therefore you can pass `.` as path to your command to reference the application directory.

Could you elaborate on the usecase of the variable? In which cases could it be needed? 
What is the advantage over passing  `.` to the command to reference the application directory?

kostyay:
> @kostyay thanks for the PR.
> 
> Tasks are always executed in the application directory also if they are defined in an include file.
> Therefore you can pass `.` as path to your command to reference the application directory.
> 
> Could you elaborate on the usecase of the variable? In which cases could it be needed?
> What is the advantage over passing `.` to the command to reference the application directory?

I will check if using . works in my usecase and update the PR :)

kostyay:
It's indeed possible to solve this without appdir, but requires some out of the box thinking. Honestly I don't think having extra variables hurts, but its up to you :).
BTW, why not support full gotemplate syntax and implement just partial limited functionality by implementing it using find/replace?
Is your idea that if I want complex logic I should extract it to an external build script?

fho:
> It's indeed possible to solve this without appdir, but requires some out of the box thinking. Honestly I don't think having extra variables hurts, but its up to you :).

I also like the idea of having an .appdir var but on the other hand it as a tiny bit more code without adding value.
The same can be already achieved without it.
Until we come up with a good usecase for the variable I'll not add it for now.


> BTW, why not support full gotemplate syntax and implement just partial limited functionality by implementing it using find/replace?

I don't have a good reason for not using gotemplates. :-)
It didn't come to my mind that I could use gotemplates when I implemented the first variables in baur 0.x.
I only had the idea quite recently and then switched the syntax in baur 1.x to be the same then for gotemplates to hopefully be able to do a smooth transition.
The priority for switching to using gotemplates wasn't high enough to do it instead of other things so far.

> Is your idea that if I want complex logic I should extract it to an external build script?

This is also true though

Issue #289, opened 2021-03-16
How to pass data from build script to `Task.Output.DockerImage.RegistryUpload`


kostyay:
Use case:
I have a build script that creates a docker image using docker build.
It decides the image tag/repo based on some parameters considered by the build script.

The docker output is configured in the following way:
```toml
        [[Task.Output.DockerImage.RegistryUpload]]

        # Registry address in the format <HOST>:[<PORT>]. If it's empty the docker agent's default is used.
        registry = ""
        repository = "{{ env DOCKER_IMAGE_REPOSITORY }}"
        tag = "{{ env DOCKER_IMAGE_TAG }}"
```

I need to somehow pass parameters into the `repository` and `tag` values as they are not constant and generated by the build script.
How can I do that? I tried setting env variables in the build script, but they are not passed down into the RegistryUpload step.

Thanks

fho:
Passing data that is generated during command execution to the upload steps is unfortunately not possible currently.
Only environment variables are accessible that were defined when the baur command was run.
The variable are also resolved before the command is executed.

I like the idea a lot, being able to pass data from the command step to the upload step would be very valuable.

kostyay:
I want to implement this as its very handy to us.
Do you have any preferred way for this to be implemented?
I was thinking perhaps build script can set env variables with a prefix and baur will parse them and make them available via helper function like env

step sets env vars:
STEP_VAR_V1
STEP_VAR_V1

inside the config can do 

{{ stepOutput "V1" }}

wdyt?

fho:
> I want to implement this as its very handy to us.

That would be great :-)


> Do you have any preferred way for this to be implemented?
> I was thinking perhaps build script can set env variables with a prefix and baur will parse them and make them available via helper function like env
[..]

If I understand it right the idea is that the executed command sets environment variables, baur reads them, stores them and later templates the upload output config with it.
Afaik it is not possible to read the environment variables that the process set.
Environment variables are available to the process and it's child processes. 
Baur executes the command as a new process, when the process terminates the environment variables that it set are gone and baur can not retrieve them.
Or am I wrong here?

Another way to pass data would be that the command writes the data to a file. The name of the file to that data must be written is defined by baur and could be made available as environment variable to the command.
When the command terminates, baur  reads and parses the file and makes the information available to the upload step.
The file could be in json format, baur unmarshals it and the datastructure can probably be made accessible via gotemplates.
We would also have to change how baur currently replaces variables (does templating).
[...]

PR #292, opened 2021-03-19
Implement go template resolver

kostyay:
Preparation for https://github.com/simplesurance/baur/issues/289
Implement go template resolver instead of the existing ones.

Do you want to keep backward compatiblity?
Go Template syntax for function calls is different than the one you implemented:
Function call is without dot and must be quoted so instead of this `{{ env Var1 }}` it must be `{{ env "Var1" }}`
I kept BC for `{{ .gitcommit }}` and `{{ .uuid }}` via your other resolver.
LMK what u think.


fho:
@kostyay 

Thanks for your effort. 

We do not need to keep backwards compatibility currently.
We can ship the change in the following 1.0-rcX version, because it's still in -rc state and will be a major release it is fine.

kostyay:
Test is failing because baur-example uses old syntax of variables.

https://github.com/simplesurance/baur-example/pull/5/files

kostyay:
@fho let me know what u think

fho:
> Test is failing because baur-example uses old syntax of variables.
> 
> https://github.com/simplesurance/baur-example/pull/5/files

The failing test is checking out the configs from baur-example in the version 0 format and is then trying to upgrade them to the new version (like running `baur upgrade configs`).
There might be a change missing in pkg/cfg/upgrade/.

I can also have a look soon and fix it if you prefer.

kostyay:
Fixed it

pkg/cfg/resolver/gotemplate.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/simplesurance/baur/v1/internal/vcs"
+)
+
+type GoTemplate struct {
+	appName     string
+	root        string
+	gitCommitFn func() (string, error)
+}
+
+func NewGoTemplate(appName, root string, gitCommitFn func() (string, error)) Resolver {
```
fho, line 21:
```suggestion
func NewGoTemplate(appName, root string, gitCommitFn func() (string, error)) *GoTemplate {
```

pkg/cfg/resolver/gotemplate.go
```diff
@@ -0,0 +1,68 @@
...
+		"appname": s.appName,
+	}
+
+	funcMap := template.FuncMap{
+		"gitcommit": func() (string, error) {
+			commit, err := s.gitCommitFn()
+			if errors.Is(err, vcs.ErrVCSRepositoryNotExist) {
+				return "", errors.New("baur repository is not part of a git repository")
+			}
+
```
fho, line 41:
```suggestion
```

removes the vcs package dependency and makes the whole package baur independent

pkg/cfg/resolver/gotemplate.go
```diff
@@ -0,0 +1,68 @@
...
+		},
+		"uuid": func() string {
+			return uuid.NewString()
+		},
+	}
+
+	t, err := template.New("baur").Funcs(funcMap).Parse(in)
+	if err != nil {
+		return "", fmt.Errorf("failed parsing go template: %w", err)
+	}
```
fho, line 60: Can we move creating templateVars, funcMap and maybe even the template to the constructor?
That way this code would be run only once instead of on every `Resolve()` call.

pkg/cfg/resolver/gotemplate.go
```diff
@@ -0,0 +1,68 @@
+package resolver
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"html/template"
```
fho, line 7: Don't we want to use `text/template` instead?
https://golang.org/pkg/text/template/

kostyay: good point

fho:
@kostyay I addressed most of the things that I mentioned.
We still have to:
- [ ] remove pkg/cfg/resolver/callback*
- [ ] remove pkg/cfg/resolver/strreplacement*
- [ ] move the resolver.Resolver interface to the cfg package
- [ ] add test for `{{ gitcommit }}` and `{{ root }}` 

and then we can merge it.

kostyay:
Hey
I didn't understand your comments correctly and force pushed over your changes 🤦 
Anyway I fixed all the comments including the ones you did, which I didn't realize were fixed. Can you compare them to your changes if you still have the branch locally? See if there anything you'd like to change?

I also fixed some tests that were failing on Mac (TempDir() on mac is symbolic link so tests were failing because tempdir returns /tmp but the search results returned /private/tmp)

kostyay:
Added table test for all gotemplate actions

fho (commented):
> I didn't understand your comments correctly and force pushed over your changes facepalm

What a pitty, next time I'll also ask before I push to your branch. :-)

> Anyway I fixed all the comments including the ones you did, which I didn't realize were fixed. Can you compare them to your changes if you still have the branch locally? See if there anything you'd like to change?

I added some more comments.
I also pushed the version from yesterday that got overwritten here: https://github.com/simplesurance/baur/tree/gotemplate_fho

pkg/cfg/resolver/gotemplate.go
```diff
@@ -0,0 +1,85 @@
...
+
+	"github.com/simplesurance/baur/v1/internal/vcs"
+)
+
+const (
+	rootVar       = "root"
+	appnameVar    = "appname"
+	gitcommitFunc = "gitcommit"
+	envFunc       = "env"
+	uuidFunc      = "uuid"
```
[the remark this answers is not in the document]
kostyay: Unfortunately gotemplate doesn't allow functions to be prefixed with dot. It panics.
```
panic: function name ".gitcommit" is not a valid identifier [recovered]
	panic: function name ".gitcommit" is not a valid identifier
```

kostyay:
Fixed all comments + Rebase over master

pkg/cfg/resolver/gotemplate.go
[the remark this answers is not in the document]
fho: ok, then we can follow the same conventions that others are using.

The widespread naming convention in gotemplate seems to be:
- variables: camelCase starting with uppercase, 
- function: camelCase starting with lowercase

For example:
- https://prometheus.io/docs/prometheus/latest/configuration/template_reference/
- https://gohugo.io/templates/introduction/

I suggest that we adopt the same convention.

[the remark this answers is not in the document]
fho: @kostyay I'll do the change

fho (approved):

fho:
@kostyay thanks a lot :tada:

PR #296, opened 2021-03-25
change version from v1 to v2

fho:
Go handles versions different then I understood.
Version 0 and version 1 of a module are expected to use the same module name
without a major version suffix.
When a "/v1" suffix is in the module name it's interpreted as part of the module
name and not as version.

This caused that baur v1 could not be installed via:
- go install github.com/simplesurance/baur@master or
- go install github.com/simplesurance/baur/v1@master

The command failed with:
	github.com/simplesurance/baur@v1.0.0-rc2: invalid version: go.mod has
	malformed module path "github.com/simplesurance/baur/v1" at revision
	v1.0.0-rc2

The error happens because https://github.com/golang/mod/blob/d6ab96f2441f9631f81862375ef66782fc4a9c12/module/module.go#L536 returns false if a module path has a '/v1' suffix.

To fix it, the major version is incremented to 2 and the module name suffix is
changed to '/v2'. For version 2 the module suffix works as expected.
A non-rc 1.0 version will not be released.
The next released version will be `2.0.0-rc1`.

Another options to fix it is to remove the '/v1' prefix from the module name,
this causes the issue that it's not possible anymore to import packages from
baur v0.20.0 because the module name is the same.
Packages from v0.20.0 are imported to upgrade configurations from version 0.20.0
to the current one.
Copying the packages instead of including them as module dependency would be an
option but we would have to duplicate multiple packages and maintain them.
This is worse.

--------

```sh
go install -x -v github.com/simplesurance/baur/v2/...@go_install
```
 works


FranciscoKurpiel (approved):
Mist! I was also not aware of it.

nocive (approved):

PR #297, opened 2021-03-25
version: initialize version variable by embedding file instead via linker flag

fho:
```
        version: improve documentation and naming

        version: initialize version variable by embedding file instead via linker flag

        When baur was compiled via it's Makefile, the content of the "ver" file was read
        and used to initialize the global Version variable of the version package.

        Instead of using a linker flag to set the variable, the ver file is
        embedded into the go binary.
        This has the advantage that no custom linker flags must be passed to set
        the version correctly.
        When baur is installed via "go install", the version will also be set
        correctly.

        The required go version in the go.mod file needed to increased because
        embedding files is only supported since go 1.16.

        The ver file was moved to the directory internal/version  because it is not
        possible to embed version from the parent directory of the package.
        A symlink was created from the old location.

        The git-commit is as before set via the linker flag. I'm not aware of an
        alternative that works without custom compile options.

```


FranciscoKurpiel (approved):
Embedded files in go! That's great! Nice use of the feature here. 🥇

Issue #300, opened 2021-04-01
Create a docker for baur


kostyay:
It would be really cool to have an official docker image to run baur.
I can make it, do you have a public repository it can be pushed to?

kostyay:
https://github.com/simplesurance/baur/pull/301

fho:
The usecases of baur differ so much that almost everybody requires a different environment with different tools
to run tasks via baur.
Some use baur to build java application, others to build node apps, some build go apps, etc.
Some others (like we internally) execute commands via baur that run in docker containers itself.

Therefore the official docker image would be only suitable for most to query information via baur but not to run any tasks.
I'm not convinced that it's worth to maintain an official image if is is only usable in such a limited scenario.

kostyay:
> The usecases of baur differ so much that almost everybody requires a different environment with different tools
> to run tasks via baur.
> Some use baur to build java application, others to build node apps, some build go apps, etc.
> Some others (like we internally) execute commands via baur that run in docker containers itself.
> 
> Therefore the official docker image would be only suitable for most to query information via baur but not to run any tasks.
> I'm not convinced that it's worth to maintain an official image if is is only usable in such a limited scenario.

Perhaps there can be a more friendly installation method then?
Maybe add homebrew support for mac or a bash file that will download and install the right version for the host machine?

fho:
Having baur packages for some linux distributions and brew would be great.
They would need to be maintained independently though.
If somebody creates a package I can link it in the README.

I would accept a PR for an sh-script that works on Linux and MacOS to install baur. :-)

thomsa:
@kostyay  I have created one some time ago. It has actually more stuff inside but unfortunately I didn't track it too much. First version has baur almost exclusively. The last version has aws-cli and nodejs, python, docker dind and many others inside. 

https://hub.docker.com/r/thomsa/baur-ci

PR #301, opened 2021-04-01
Add dockerfile and make commands

kostyay:
https://github.com/simplesurance/baur/issues/300


fho:
https://github.com/simplesurance/baur/issues/300#issuecomment-814307184

PR #305, opened 2021-04-19
support to pass multiple --input-str parameters

fho:
The run, status, ls inputs, diff inputs commands now support to be passed multiple
"--input-str" parameters, each specified parameter is added as individual string
input.

### DIFF
--- a/internal/command/command_test.go
+++ b/internal/command/command_test.go
@@ -55,6 +55,21 @@ func baurCSVLsApps(t *testing.T) [][]string {
 	return statusOut
 }
 
+// baurCSVLsInputs runs "baur ls inputs --csv" and returns the list of files as
+// string slice. args is passed as argument to the "baur ls inputs" command.
+func baurCSVLsInputs(t *testing.T, args ...string) []string {
+	t.Helper()
+
+	stdoutBuf, _ := interceptCmdOutput(t)
+
+	lsInputsCmd := newLsInputsCmd()
+	lsInputsCmd.csv = true
+
+	lsInputsCmd.Command.Run(&lsInputsCmd.Command, args)
+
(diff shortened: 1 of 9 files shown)


FranciscoKurpiel (commented):
If I understand correctly, the code as it is, will produce the same final hash for the inputs if I just invert the order of the `--input-str parameters` flags. For example:

```
--input-str A --input-str B
```

and

```
--input-str B --input-str A
```

fho:
@FranciscoKurpiel yes that is correct

FranciscoKurpiel:
> @FranciscoKurpiel yes that is correct

Imho, the order should be preserved, because I believe most frequently than not, it will hold semantic value, so if we invert the order the artifact should be rebuilt. One possible approach for that is instead of this additional inputs being just strings, they could be key-pair maps. For example:

```
--input-str branch=master --input-str mode=develop
```

fho:
@FranciscoKurpiel 

It's possible to use `--input-strings` in both ways you mentioned because they are arbitrary and can be freely defined:
If you want use them in a way where the order matters, you could specify them as one input-string instead of multiple, like: 
`--input-str branch=master,mode=develop`.
If you want to use a key-value format, you could always specify them in a kv format, like `--input-str branch=master --input-str mode=develop`.

I think it would be counter-intuitive and would led to unexpected results if the order in which the `--input-strings` parameters are specified on the commandline would matter.
You would always have to remember in which order they were specified and ensure you specify them in the same order in all invocations, otherwise you get unexpected results

FranciscoKurpiel:
Fair points. Would you like to proceed with this design or will you tweak more?

fho:
@FranciscoKurpiel this PR is done :-)

FranciscoKurpiel (approved):

PR #306, opened 2021-04-19
storage: create unique input table index over URI md5 hash

fho:
When very long "--input-strings" were passed to the "baur run" command, storing
the run in the database failed with:
	index row requires 10032 bytes, maximum size is 8191 (SQLSTATE 54000)

The input table has a unique constraint for the URI and digest field.
When very long input strings are used the 8181B limit for the size of a column
in the index is reached.

Use the md hash of the URI in the index instead of it's raw string value. This
prevents that we hit the limit.

Because only a -rc version was released for this version, no migrations will be
provided.
A 2.0.0-rc1 db can be migrated to use the new index manually via:
	ALTER TABLE input DROP CONSTRAINT input_uri_digest_uniq;
	CREATE UNIQUE INDEX input_uri_digest_uniq ON input (MD5(uri), digest);

### DIFF
--- a/pkg/storage/postgres/insert.go
+++ b/pkg/storage/postgres/insert.go
@@ -152,7 +152,7 @@ func insertInputIfNotExist(ctx context.Context, db dbConn, inputs []*storage.Inp
 	   VALUES
 `
 	const stmt2 = `
-	       ON CONFLICT ON CONSTRAINT input_uri_digest_uniq
+	       ON CONFLICT (MD5(uri), digest)
 	       DO UPDATE SET id=input.id
 	RETURNING id
 	`
--- a/pkg/storage/postgres/schema.go
+++ b/pkg/storage/postgres/schema.go
@@ -65,9 +65,10 @@ CREATE TABLE task_run (
 CREATE TABLE input (
 	id serial PRIMARY KEY,
 	uri text NOT NULL,
-	digest text NOT NULL,
-	CONSTRAINT input_uri_digest_uniq UNIQUE (uri, digest)
+	digest text NOT NULL
 );
+
+CREATE UNIQUE INDEX input_uri_digest_uniq ON input (MD5(uri), digest);
 
 CREATE TABLE task_run_input (
 	task_run_id integer NOT NULL REFERENCES task_run(id) ON DELETE CASCADE,


FranciscoKurpiel (commented):
I'm not convinced it is a good approach to use the db-side md5 function. We are already using sha-384 for other resources, I think it makes sense to use the same for this.

What about having two columns, `uri` and `urlhash`, and using the `(urihash, digest)` as the unique constraint? The `urihash` value would never be exposed outside of the db code, or course.

fho:
> I'm not convinced it is a good approach to use the db-side md5 function. We are already using sha-384 for other resources, I think it makes sense to use the same for this.

I don't see an advantage of using the same hash then somewhere else here.
For the hash function it is important that it is unlikely that for the URI the same digest is generated and that it is not too slow.
The result of my google search was that it is very unlikely that the same md5 hash for different data will be generated and MD5 hashes are still acceptable for this usecase.  Would you agree with that? :-)
Performance-wise both would be fine, MD5 is faster then SHA but the noticeable difference should be so small in this scenario that it does not matter.

The biggest advantage of using MD5 here is that it works out of the box with postgres.
If we want to use SHA hash functions, the pgcrypto extension must be installed and enabled explicitly for the database.
Some Linux distributions ship the extension in a separate package then the psql server.



> What about having two columns, `uri` and `urlhash`, and using the `(urihash, digest)` as the unique constraint? The `urihash` value would never be exposed outside of the db code, or course.

What would be the advantage if we would have an urlhash column instead of letting psql manage creating and storing the digests in the index table?

FranciscoKurpiel:
> I don't see an advantage of using the same hash then somewhere else here.

The two advantages are collision resistance and consistency. MD5 is generally considered outdated because its collision resistance properties are not on pair with more modern algorithms. Even sha1 is considered not outdated.

> For the hash function it is important that it is unlikely that for the URI the same digest is generated and that it is not too slow.

Fair point.

> The result of my google search was that it is very unlikely that the same md5 hash for different data will be generated and MD5 hashes are still acceptable for this usecase. Would you agree with that? :-)

I think it is acceptable. I would prefer a sha2-based method, but md5 is unlikely to cause us issues.

> Performance-wise both would be fine, MD5 is faster then SHA but the noticeable difference should be so small in this scenario that it does not matter.

Agree.

> The biggest advantage of using MD5 here is that it works out of the box with postgres.
> If we want to use SHA hash functions, the pgcrypto extension must be installed and enabled explicitly for the database.
> Some Linux distributions ship the extension in a separate package then the psql server.

I read about pgcrypto, and I don't like it.

> What would be the advantage if we would have an urlhash column instead of letting psql manage creating and storing the digests in the index table?

Not depending on pgcrypto and being able to use sha384.

I think the current approach is reasonable. Even though I would prefer the trade offs of using sha384, I admit that the advantages are not big, so I'm totally fine merging it like this.

fho:
issues is fixed by PR https://github.com/simplesurance/baur/pull/306

PR #309, opened 2021-04-22
storage: refactor input string representation

fho:
```
        tests: add testcase "baur ls runs --has-input"

        storage: refactor input string representation

        File Input and String Inputs were represented by the same storage struct
        and also stored in the same postgresql table.
        In the input.name db column and in the storage.Input.Name struct field was
        stored for file inputs the path, for string inputs the string with a "string:"
        prefix.
        The input table had a unique constraint for the name and digest column, that was
        used to to a database upsert operation (only insert if records does not exist).
        When long string inputs were specified as parameters for "baur run" it happened
        that the input could not be stored in the table because it was too long to be
        stored in the unique index (index row requires 10032 bytes, maximum size is 8191
        (SQLSTATE 54000)).

        To fix this issue and to have clear struct field and db column names the storage
        representation is refactored.
        Input Files and Input strings are now represented by separate structs and stored
        in separate tables.
        The unique index for the input_string table is only created for the digest field
        and not the string itself anymore. The digest depends only on the string.

```

### DIFF
--- a/pkg/storage/postgres/query.go
+++ b/pkg/storage/postgres/query.go
@@ -47,18 +47,17 @@ func (c *Client) LatestTaskRunByDigest(ctx context.Context, appName, taskName, t
 	       task.name,
 	       vcs.revision,
 	       vcs.dirty,
-	       task_run_input.total_digest,
+	       task_run.total_input_digest,
 	       task_run.start_timestamp,
 	       task_run.stop_timestamp,
 	       task_run.result
 	  FROM application
 	  JOIN task ON application.id = task.application_id
 	  JOIN task_run ON task.id = task_run.task_id
-	  JOIN task_run_input ON task_run_input.task_run_id = task_run.id
 	  LEFT OUTER JOIN vcs ON vcs.id = task_run.vcs_id
 	 WHERE application.name = $1
 	   AND task.name = $2
-	   AND task_run_input.total_digest = $3
+	   AND task_run.total_input_digest = $3
 	 ORDER BY task_run.stop_timestamp DESC
 	 LIMIT 1
 	 `
@@ -89,16 +88,16 @@ func (c *Client) LatestTaskRunByDigest(ctx context.Context, appName, taskName, t
 	return &result, nil
 }
 
-func (c *Client) Inputs(ctx context.Context, taskRunID int) ([]*storage.Input, error) {
+func (c *Client) inputStrings(ctx context.Context, taskRunID int) ([]*storage.InputString, error) {
 	const query = `
-	SELECT input.uri,
-	       input.digest
-	  FROM input
-	  JOIN task_run_input ON input.id = task_run_input.input_id
-	  WHERE task_run_input.task_run_id = $1
-	  `
+	SELECT input_string.string,
+	       input_string.digest
+	  FROM input_string
+	  JOIN task_run_string_input ON input_string.id = task_run_string_input.input_string_id
+	 WHERE task_run_string_input.task_run_id = $1
+	 `
 
-	var result []*storage.Input
+	var result []*storage.InputString
 
 	rows, err := c.db.Query(ctx, query, taskRunID)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


pkg/storage/postgres/schema.go
```diff
@@ -56,28 +56,46 @@ CREATE TABLE task_run (
...
 
-CREATE TABLE input (
+CREATE TABLE input_file (
 	id serial PRIMARY KEY,
-	uri text NOT NULL,
+	path text NOT NULL,
+	digest text NOT NULL,
+	CONSTRAINT input_file_path_digest_uniq UNIQUE (path, digest)
+);
+CREATE INDEX idx_input_file_path ON input_file(path);
```
FranciscoKurpiel, line 73: Do we need this index?

pkg/storage/postgres/schema.go
```diff
@@ -56,28 +56,46 @@ CREATE TABLE task_run (
...
+
+CREATE TABLE input_string (
+	id serial PRIMARY KEY,
```
FranciscoKurpiel, line 76: optional: since we use `digest` on an unique constraint, we could use the digest as the primary key and get rid of the serial column.

pkg/storage/postgres/query.go
```diff
@@ -89,16 +88,55 @@ func (c *Client) LatestTaskRunByDigest(ctx context.Context, appName, taskName, t
...
 
-func (c *Client) Inputs(ctx context.Context, taskRunID int) ([]*storage.Input, error) {
+func (c *Client) inputStrings(ctx context.Context, taskRunID int) ([]*storage.InputString, error) {
 	const query = `
-	SELECT input.uri,
-	       input.digest
-	  FROM input
-	  JOIN task_run_input ON input.id = task_run_input.input_id
-	  WHERE task_run_input.task_run_id = $1
+	SELECT input_string.string,
```
FranciscoKurpiel, line 93: Input strings are not guaranteed to be returned always on the same order, and the final hash of the input will change. I suggest adding `ORDER BY input_string.digest`.

FranciscoKurpiel (commented):
Nice! So you decided to store the full input string on a non-indexed column and the digest on a indexed one. I like the approach quite a bit!

Please take a look at the `order by` point, The rest is optional.

pkg/storage/postgres/schema.go
fho: `baur ls runs --input-string=PATH` searches in the table for a given path, the index should speed up those look ups.

Same situation  for input string + `baur ls runs --input-string=string:hello` but we cannot have an index on the string column.
The command+query can be changed  for input-strings instead to create the digest for the string and query query for the indexed digest in the table instead. The change can be done if that will ever become an issue.

pkg/storage/postgres/schema.go
fho: Yes, I'm using a numerical PK because it only uses 4 Bytes whenever it is stored as FK in other tables. 
I also expect that queries are faster when the FK is only 4B, I don't know that for sure though :-)

pkg/storage/postgres/query.go
fho: That is fine, there is currently notrequirement for returning inputs ordered from the storage.
The CLI commands are sorting inputs/outputs before they are printed to the console

FranciscoKurpiel: I keep forgetting that you don't use the file and input string digest to reconstruct the final digest for the inputs.

FranciscoKurpiel (approved):

pkg/storage/postgres/schema.go
fho: Good points! I was not considering those.

> For searches, the current approach requires to always join the the table on a query, even if you only want the digest

Right, currently we do not have this usecase in baur.

> Only searches that read the text of the string parameter will be improved by the smaller size of the column used on the relationship.

I would expect also Joins because the key is much smaller.
I have no clue what differences that does in practice, probably not noticeable until  there are giant amounts of records in the tables  :-)

I will keep it as it is for now.

Issue #303, opened 2021-04-05
Running build jobs in parallel


kostyay:
Hey
Is there any reason why the builds at the moment are synchronous?
Do you see any architectural challenges making builds run in go routine pools?
Thanks

fho:
The only reason is that it did not seem worth to implement for our specific usecase so far. :-)
I think it would be great if baur would support a commandline option to configure the number of tasks to run in parallel.
Most of what is needed for it already exist.
The only tricky part to consider is how we handle the output of the tasks in `--verbose` mode. 
Either baur shows output of tasks run in parallel interleaved, then it becomes hard to follow or baur buffers the output of parallel executed tasks and shows it when they terminated. If a task hangs, baur won't show it's output until it terminated which will make it harder to analyze the issues.

If baur is used to run tasks that all have a similar workload or on machines that have much more resources available then the average task requires, running them in parallel should be much more efficient.

When baur runs tasks with mixed workloads on multiple machines with low resources it seems impossible to find an efficient limit for the number of parallel executions though.
Some tasks require only minimal resources others tasks utilize the whole machine.
If multiple tasks are run in parallel that each could utilize the whole machine, running them in parallel will be slower in total then running them sequentially. It might even cause OOM issues.

In our setup we currently parallelize and distribute task execution on multiple machines via Jenkins:
- run `baur status` to figure out which tasks needs to be run,
- run `baur run TASK` for each tasks in parallel via jenkins, jenkins schedules them across the available executors on multiple nodes

This has the overhead of discovering and calculating the digest for the same task 2x, though.

kostyay:
I realize running tasks in parallel could be counter productive and inefficient, however the ability I want is similiar to running `make -j <number>`. 
```       -j [jobs], --jobs[=jobs]
            Specifies  the  number  of  jobs (commands) to run simultaneously.  If there is more than one -j option, the last one is effective.  If the -j option is given
            without an argument, make will not limit the number of jobs that can run simultaneously.```

fho:
awesome, thanks :-)

kostyay:
See PR

fho:
PR was merged, thanks for your help @kostyay

PR #317, opened 2021-04-23
Support running build tasks simultaneously

kostyay:
Fixes issue #303


internal/command/run.go
```diff
@@ -106,6 +108,8 @@ func newRunCmd() *runCmd {
 		"include a string as an input")
 	cmd.Flags().StringVar(&cmd.lookupInputStr, "lookup-input-str", "",
 		"if a run can not be found, try to find a run with this value as input-string")
+	cmd.Flags().UintVarP(&cmd.buildWorkers, "tasks", "t", 1,
```
fho, line 111: Please validate if a valid value was passed for the parameter and otherwise fail with an error.
Running `baur run -t 0` would currently hang forever

internal/command/run.go
```diff
@@ -106,6 +108,8 @@ func newRunCmd() *runCmd {
...
+		"specifies  the  number  of  tasks (builds) to run simultaneously")
```
fho, line 112: something weird happened here with spaces :-)

internal/command/run.go
```diff
@@ -122,6 +126,7 @@ func (c *runCmd) run(cmd *cobra.Command, args []string) {
 	defer c.storage.Close()
 
 	c.uploadRoutinePool = routines.NewPool(1) // run 1 upload in parallel with builds
+	c.buildRoutinePool = routines.NewPool(c.buildWorkers)
```
fho, line 129: in the master branch "builds" are replaced with the concept of tasks, the pool could be named `taskRunnerRoutinePool` instead

fho (commented):
cool, thanks a lot for the PR :+1:

Could you also add a testcase that runs tasks in parallel? At best it also checks if they are really executed in parallel.

kostyay:
> cool, thanks a lot for the PR 👍
> 
> Could you also add a testcase that runs tasks in parallel? At best it also checks if they are really executed in parallel.

It's not very simple. I will think how to do it, any suggestions?
I fixed the other comments.

fho:
> > cool, thanks a lot for the PR +1
> > Could you also add a testcase that runs tasks in parallel? At best it also checks if they are really executed in parallel.
> 
> It's not very simple. I will think how to do it, any suggestions?

The tasks could run shell-scripts that check if another process with the name of the other process is running.
If it is they exit with code `0` otherwise with code `1`.
I don't know what is available on the Windows image that is used in CI though, `ps` is probably not available.

Alternatively the tasks could write a timestamp to a file after they started, sleep for some seconds, write another timestamp to the file before they terminated.
The testcase reads the files and checks if the 2 timeranges overlap.
That should also work on Windows, bash's `printf` could be used to write the timestamp (`printf "%(%s)T > file"`)

kostyay:
> > > cool, thanks a lot for the PR +1
> > > Could you also add a testcase that runs tasks in parallel? At best it also checks if they are really executed in parallel.
> > 
> > 
> > It's not very simple. I will think how to do it, any suggestions?
> 
> The tasks could run shell-scripts that check if another process with the name of the other process is running.
> If it is they exit with code `0` otherwise with code `1`.
> I don't know what is available on the Windows image that is used in CI though, `ps` is probably not available.
> 
> Alternatively the tasks could write a timestamp to a file after they started, sleep for some seconds, write another timestamp to the file before they terminated.
> The testcase reads the files and checks if the 2 timeranges overlap.
> That should also work on Windows, bash's `printf` could be used to write the timestamp (`printf "%(%s)T > file"`)

good idea, i will see what i can come up with over the weekend

fho:
@kostyay I will try to pick-up the PR tomorrow and can then also add the testcases

kostyay:
Cool thanks.
Didn't have time to get to it

fho:
Done, I merged the branch manually to [master](https://github.com/simplesurance/baur/commits/master)

PR #324, opened 2021-04-30
loader: fix: a task spec with only a wildcard for the app loaded all tasks

fho:
```
  loader: fix: a task spec with only a wildcard for the app loaded all tasks

        If a task specification was passed to "baur status" or "baur run" that consisted
        of a wildcard for the app and a specific task-name like '*.build', all tasks for
        all apps were loaded,  instead of only all tasks matching the task-name.

        Fix it by removing the specs.allApps fields.
        It is not necessary because if a spec is parsed with an non-empty task part an
        element for it is added to specs.taskSpecs and it will then be loaded by
        Loader.tasks().

        loader: rename appTask() to appTaskByName()

        The new name should be more clear

        loader: improve debug log message

        - return early in appNames() when no names were passed.
          Passing no names is supported behavior and used in multiple places, so far the
          function was nothing because all code blocks were run in for-loops over an
          empty slice in the case.
          Adding an if-check at the beginning makes it easier to figure out what happens
          in the case and prevents that we log a debug message about loading no apps by
          name

        testcases: improve status command testcase

        - verify in the testcases if the expected tasks were returned from "baur status"
        - add testcase that passes multiple different testspecs to "baur status"
        - improve testcase names

        tests: interceptCmdOutput: reset stdout, stderr when test finishes

        This fixes the same issue then 7b76d65, when the following testcase did not
        call interceptCmdOutput() or initTest() the output was written to the logger of
        a previous testcase and vanished.
```

### DIFF
--- a/internal/command/status_test.go
+++ b/internal/command/status_test.go
@@ -3,7 +3,6 @@
 package command
 
 import (
-	"fmt"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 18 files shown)


internal/command/status_test.go
```diff
@@ -13,93 +12,208 @@ import (
...
+				t.Cleanup(func() {
+					require.NoError(t, os.RemoveAll(childDir))
+				})
+
+				cwd, err := os.Getwd()
+				require.NoError(t, err)
+
+				require.NoError(t, os.Chdir(childDir))
+				t.Cleanup(func() {
+					require.NoError(t, os.Chdir(cwd))
```
FranciscoKurpiel, line 173: Cleanup on a preparation function, very nice trick! I'm totally staling this idea.

FranciscoKurpiel (approved):

PR #332, opened 2021-06-10
readme: add instructions for building and installing baur from source

fho:


nocive (approved):
i can testify it doesn't work with go 1.15.1 :sweat_smile:

PR #346, opened 2022-03-28
postgres: prevent db deadlock on input insert with multiple instances

fho:
When multiple baur instances ran in parallel and were inserting inputs it
happened that a postgresql deadlock occurred.
The happened deadlock is exactly the one explained in
https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-DEADLOCKS.

To prevent those deadlocks, the inputs are now sorted before insert.
This prevents the described deadlock scenario.

This fixes the postgresql error:
	2022-02-25 09:37:26 UTC ERROR:  deadlock detected
	2022-02-25 09:37:26 UTC DETAIL:  Process 158747 waits for ShareLock on
					 transaction 44098; blocked by process
					 157931.
		Process 157931 waits for ShareLock on transaction 44099; blocked
		by process 158747.
        Process 158747:
                   INSERT INTO input_file (path, digest)
	[...]

### DIFF
--- a/pkg/storage/postgres/insert.go
+++ b/pkg/storage/postgres/insert.go
@@ -3,6 +3,7 @@ package postgres
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/jackc/pgx/v4"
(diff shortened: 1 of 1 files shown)


pkg/storage/postgres/insert.go
```diff
@@ -144,14 +144,16 @@ func insertVCSIfNotExist(ctx context.Context, db dbConn, revision string, isDirt
 
 func insertInputFilesIfNotExist(ctx context.Context, db dbConn, inputs []*storage.InputFile) ([]int, error) {
 	const stmt1 = `
-           INSERT INTO input_file (path, digest)
-	   VALUES
-`
+	     WITH data (path, digest) AS (
```
FranciscoKurpiel, line 147: Isn't this exactly the same as sort sorting `inputs` by `inputs.Path` and `inputs.Digest` on go, then execute the sql statement as before? I think the code would read a lot better, and the effect on avoiding the deadlock would be the same, no? I mean, this code looks totally correct, but it took me a few minutes to understand it.

pkg/storage/postgres/insert.go
```diff
@@ -177,10 +179,13 @@ func insertInputFilesIfNotExist(ctx context.Context, db dbConn, inputs []*storag
 
 func insertInputStringFilesIfNotExist(ctx context.Context, db dbConn, inputs []*storage.InputString) ([]int, error) {
 	const stmt1 = `
-           INSERT INTO input_string (string, digest)
-	   VALUES
-`
+	     WITH data (string, digest) AS (
```
FranciscoKurpiel, line 182: Same here.

pkg/storage/postgres/insert.go
fho: Yes, it is exactly the same and I agree, it's easier to understand when sorting the slice in Go.
Performance-wise it should also be fine. The additional required memory of the slice copy also does not matter much, it only contains pointers.
I changed it.

thanks for the suggestion!

FranciscoKurpiel (approved):

PR #356, opened 2022-09-08
ci: use github actions instead of circleci

fho:
Running the tests on Windows CircleCI machines was was slowly.
Mainly the installation process took a long time, the github action windows
image already have everything installed what is needed (psql, gcc).

Switch to using github actions for CI.
The CI jobs now also run the tests with CGO disabled.
When CGO is disabled using the race detector is not possible. The release are
build with CGO disabled, therefore we also run the test without it again.

In CI currently only the Linux baur binary is build. This should be sufficient.
I never had the case that I could compile baur on linux but not for another
architecture or OS.


fho:
The failing windows testcase will be fixed as part of https://github.com/simplesurance/baur/issues/358

PR #365, opened 2022-09-14
fs/FindFileInparentDirs: ensure search ends on root path os independent

fho:
address the Todo note that FindFileInparentDirs() might not have been able to detect the root path on all OSes.
This is achieved by:
- using filepath.Dir() to get the path to the parent directory instead of using "..", which might also not have been worked on all OSes. filepath.Dir() returns a path ending in os.PathSeparator only when it is the root path, this is used to detect the root path.
- To prevent that the search aborts too early if an unclean path ending in a PathSeperator is passed as startPath, filepath.Clean() is called on the startPath now


FranciscoKurpiel (approved):
Interesting approach for finding out what the root path is.

PR #369, opened 2022-09-27
storage: support applying schema migrations

fho:
```
storage: support applying schema migrations

This commit adds support to change the database schema via migrations.

Schema migrations are stored in the embedded FS pkg/storage/postgres/migrations.
They are named as <VERSION>.sql. <VERSION> must be an int32 >=1.
The initial db schema that was stored as a constant in schema.go is moved to
pkg/storage/postgres/migrations/1.sql.

The migrations files are read and the filename is parsed when Init() or
ApplyMigrations() is called.
Init() applies all found migrations files.
ApplyMigrations() only applies the migrations that have a version that is
smaller then schema version stored in the database.

ApplyMigrations() will be used in a following commit.

storage/insert: use BeginFunc from pgx

Use the handy BeginFunc from pgx to use a transaction in SaveTaskRun, instead of
implementing the Begin/Commit/Rollback logic on our own.

```

### DIFF
--- a/pkg/storage/postgres/client.go
+++ b/pkg/storage/postgres/client.go
@@ -53,6 +53,7 @@ func (c *Client) Close() error {
 }
 
 type dbConn interface {
+	BeginFunc(context.Context, func(pgx.Tx) error) error
 	QueryRow(context.Context, string, ...interface{}) pgx.Row
 	Query(context.Context, string, ...interface{}) (pgx.Rows, error)
 	Exec(context.Context, string, ...interface{}) (pgconn.CommandTag, error)
--- a/pkg/storage/postgres/schema_test.go
+++ b/pkg/storage/postgres/schema_test.go
@@ -50,3 +50,30 @@ func TestIsCompatible_SchemaVersionDoesNotMatch(t *testing.T) {
 	err = client.IsCompatible(ctx)
 	assert.Error(t, err, "database schema version is not compatible")
 }
+
+func TestApplyMigrations(t *testing.T) {
+	client, cleanupFn := newTestClient(t)
+	defer cleanupFn()
+
+	require.NoError(t, client.Init(ctx))
+
+	err := client.ApplyMigrations(ctx, []*migration{
+		{
+			version: 1,
+			sql:     "CREATE table t1()",
+		},
+		{
+			version: 2,
+			sql:     "CREATE table t2()",
+		},
+	})
+	require.NoError(t, err)
+
+	exist, err := client.tableExists(ctx, "t1")
+	require.NoError(t, err)
+	require.True(t, exist, "t1 table does not exist")
+
+	exist, err = client.tableExists(ctx, "t2")
+	require.NoError(t, err)
+	require.True(t, exist, "t2 table does not exist")
+}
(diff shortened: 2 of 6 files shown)


pkg/storage/postgres/client.go
```diff
@@ -53,6 +53,7 @@ func (c *Client) Close() error {
 }
 
 type dbConn interface {
+	BeginFunc(context.Context, func(pgx.Tx) error) error
```
FranciscoKurpiel, line 56: This looks a bit inconsistent to me. We can use this interface to create transaction, but the other methods execute outside of transactions. I would like to propose an approach for this:

```go
type txMan interface {
  BeginTx() tx
}

type tx interface {
  QueryRow(context.Context, string, ...interface{}) pgx.Row
  Query(context.Context, string, ...interface{}) (pgx.Rows, error)
  Exec(context.Context, string, ...interface{}) (pgconn.CommandTag, error)
  Commit() error
  Rollback() error
}
```

If adopted, the `storage` package would be changed to instead of having a `Storer` interface, having instead a `TxManager` and a `Tx` interface. All requests would go through a transaction, and all methods would be changed to receive a transaction instead of a `Storer`. This would have the (imo) huge advantage of not leaving garbage behind when storing runs and other objects that use multiple records on multiple tables.

pkg/storage/postgres/schema_test.go
```diff
@@ -50,3 +50,30 @@ func TestIsCompatible_SchemaVersionDoesNotMatch(t *testing.T) {
...
+func TestApplyMigrations(t *testing.T) {
+	client, cleanupFn := newTestClient(t)
+	defer cleanupFn()
+
+	require.NoError(t, client.Init(ctx))
+
+	err := client.ApplyMigrations(ctx, []*migration{
+		{
+			version: 1,
+			sql:     "CREATE table t1()",
```
FranciscoKurpiel, line 63: Column-less tables on their best! :heart:

FranciscoKurpiel (commented):
Nice!

I would like to debate the transaction topic before approving.

pkg/storage/postgres/client.go
fho: The [storer interface](https://github.com/simplesurance/baur/blob/1408113e5cb3f698ec024868f4cf679f2b363c7a/pkg/storage/storage.go#L93) does not have to be aware of transactions. Transactions are imho an implementation detail of this level. The methods of this interface should ensure that data is consistent, however this is done internally.
All operations that store data (it's actually only one https://github.com/simplesurance/baur/blob/1408113e5cb3f698ec024868f4cf679f2b363c7a/pkg/storage/postgres/insert.go#L510-L517) are already using transactions.

Does this make sense or do you have something else in mind?

FranciscoKurpiel: The use case to which I was proposing the use of a transaction manager don't actually match baur very well, since the storage interface have "big methods", that receive complete objects and store them atomically. The transaction manager would fit the case where there are smaller methods, and where multiple calls have to be done consistently, withing the same transaction. Because of that I will withdraw my suggestion.

FranciscoKurpiel (approved):

PR #371, opened 2022-10-11
inputs: support tracking environment variables

fho:
This commits adds support to track the value of environment variables as Input. As all other inputs, they can be defined in an include or app configuration file.
The .app.toml configuration sections looks like:
```toml
  [[Task.Input.EnvironmentVariables]]
    names = [
      "APP_VERSION",
      "PAYMENT_APP_*",
    ]
    optional = false
```

Environment variables names can contain go glob patterns. This allows to track all variables that a certain patterns without having to list each of them. If the optional flag is true, baur command that calculate app inputs (baur status, baur run, etc) will fail if no environment variable is defined matching the defined pattern.
Declared but undefined environment variables are treated as non-existing.

The values of environment variables is not stored in the database or shown by "ls inputs", to prevent that secrets are accidentally shown or stored. A new database migration is added that adds table to store the digests of environment variables of runs.

Demonstration: https://github.com/simplesurance/baur-example/commit/2800bb04437aa666a7c76b2beffa41a91b874911

### DIFF
--- a/internal/command/run_test.go
+++ b/internal/command/run_test.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/simplesurance/baur/v2/internal/testutils/repotest"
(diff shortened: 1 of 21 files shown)


FranciscoKurpiel (commented):
I don't understand the motivation for having to list the environment variables that are being traced. To me it looks like it would make sense to either:
- track all env vars -- or
- track all env vars, but allow blacklisting some of them, then sanitizing the environment when invoking the command by removing the blacklisted vars -- or
- track all environment variables by default, but allow specifying a whitelist of variables; the environment of the build command is then sanitized by removing variables not on the list

The approach looks fragile, because variables like `PATH`, `CCACHE`, or any other env var that may affect the build command and the resulting artifacts may be missed.

Am I'm missing something?

The approach

fho:
@FranciscoKurpiel 

> I don't understand the motivation for having to list the environment variables that are being traced. 

Imagine you have a task that uses an environment variable called GIT_COMMIT. Whenever the value changes the task execution can have a different result.
With this PR it will be possible to rerun the task if the file inputs or the environment variable changed.

>To me it looks like it would make sense to either:
> 
>     track all env vars -- or
>     track all env vars, but allow blacklisting some of them, then sanitizing the environment when invoking the command by removing the blacklisted vars -- or

Tracking more environment variables then needed does not work.
It means that:
- tasks are rerun despite an environment variable can not change the output,
- you do not get the same results when running "baur status", because local environment variables that have no 
- influence on a task have different values.

> track all environment variables by default, but allow specifying a whitelist of variables; the environment of the build command is then sanitized by removing variables not on the list

Tracking all and having to blacklist all variables that do not matter does not scale.
Imagine we want to run "baur status" on mine and your machine and get the same result.
I have currently 87 environment variables defined in my shell.
In yours different ones are assigned. When you want to test something locally you might define a new variable in your shell and forget to undefine it, running "baur status" afterwards would give a different result.
You would have to blacklist all possible env variable that do not matter. This does not scale.


> The approach looks fragile, because variables like PATH, CCACHE, or any other env var that may affect the build command and the resulting artifacts may be missed.

That is true, you might miss to track some environment variables. It is the same issue that we have with tracking files.
Tracking more variables then needed is worse. The result would be that tasks are rerun despite they are producing the same output. It would not be possible to run "baur status" on multiple machines and figure out which tasks already run on a commit, because the environment differs everywhere.


> Am I missing something?

The task executor can be changed in the future, to run commands with an environment where only the tracked variables are available.

FranciscoKurpiel (approved):

PR #374, opened 2022-10-25
implement "baur upgrade db" command

fho:
Implement a new baur subcommand (`baur upgrade db`) to upgrade the database schema to a newer version.
The public storage API is extended for methods to upgrade db and get the current and required schema version.
Testcases for the `upgrade db` command are added, which are creating a database schema via the baur 2.x packages and then upgrading the db.

See commit messages for more details


FranciscoKurpiel (approved):
Interesting approach, using the `v2` to create DB with old schema, to test the migrations. :+1:

PR #378, opened 2022-11-08
inputs: support to exclude files by glob patterns

fho:
inputs: support to exclude files by glob patterns

Support to exclude files from the inputs by specifying them in an ignore list.
The list of files to exclude can be specified as glob patterns in an
"Input.ExcludedFiles" section. Patterns in the ExcludedFiles section that
reference non-existing paths are silently ignored.
The ExcludedFiles are processed after all other input sections.
Input files that were discovered and match one of the ExcludeFiles patterns are
removed.
ExcludedFiles sections can be defined in include files and imported on task
level.

In the .toml config file the Files, GolangSources and EnvironmentVariables
members of the input sections are lists.
E.g. '[[Task.Input.Files]]' sections for a task can exit with different options.
The ExcludedFiles is a single element, because it does not have additional
options.
Because it is not a list, the representation in the toml config differs:
'[Task.Input.ExcludeFiles]' vs '[[Task.Input.Files]]'

Example .app.toml section:

```toml
  [Task.Input.ExcludedFiles]
    paths = ["dbmigrations/1.sql"]

```

### DIFF
--- a/pkg/baur/inputresolver.go
+++ b/pkg/baur/inputresolver.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/simplesurance/baur/v3/internal/fs"
 	"github.com/simplesurance/baur/v3/internal/log"
 	"github.com/simplesurance/baur/v3/internal/resolve/glob"
 	"github.com/simplesurance/baur/v3/internal/resolve/gosource"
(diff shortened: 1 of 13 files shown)


pkg/baur/inputresolver.go
```diff
@@ -202,6 +220,17 @@ func (i *InputResolver) pathsToUniqInputs(repositoryRoot string, paths []string)
 
 		dedupMap[path] = struct{}{}
 
+		excluded, err := i.globPathResolver.Matches(path, excludePatterns)
+		if err != nil {
+			return nil, fmt.Errorf("ExcludedFiles: %w", err)
+		}
+
+		if excluded {
+			// TODO: log the pattern that caused it's exclusion
```
nocive, line 229:
```suggestion
			// TODO: log the pattern that caused its exclusion
```

nocive (approved):
I just have my usual contribution :sweat_smile: 
Jokes aside, it looks great, looking forward to leverage the new capabilities.

pkg/baur/inputresolver.go
fho: Thanks for reminding me about that TODO comment. :-)
I implemented the missing part + removed the comment.

nocive: glad i could be of any use :sweat_smile:

FranciscoKurpiel (approved):

PR #380, opened 2022-11-09
support shell completion for zsh, fish and powershell

fho:
Add support to generate shell completion scripts for zsh, fish and powershell.

See the commit messages for details.


FranciscoKurpiel (approved):
:100:

PR #382, opened 2022-11-09
 shell completions: support arg completion 

fho:
This PR adds support for completing arguments for all baur commands.

If the baur command is run in a baur repository, the configuration files are passed and the available apps, tasks and directories containing apps are used for completion.
Completing to arguments that must be fetched from the database, e.g. run-ids or task-id for `ls runs` is not implemented in this PR.

```
completions: complete "ls runs" args to app and task-ids

Support to complete argument pass to "ls runs" to task-ids.

completions: disable file completions for "ls outputs"

"ls outputs" only supports database IDs as argument, do not offer completions to
filenames and directories.

completions: support argument completion for "ls inputs"

Support to complete "ls inputs" arguments to task-ids and app-directories,
completing to run-ids is not supported.

completions: support completing "diff inputs" arguments

Support to complete "diff inputs" arguments to task ids found in the repository.
Arguments are not completed to task ids found in the repository, to run-ids and
the '^' syntax.

completions: support completing "baur show" arguments

Add support for completing the argument of "baur show".

Run-IDs are currently not completed.
It would be possible, by querying the existing IDs from the database that start
with the same numbers then the ones on the commandline. If a lot of records
exist, doing e.g. "baur show 1<TAB>" results in thousands of options.
That does not seem very helpful nor necessary.

ls outputs: fail earlier when passed argument is not a number

If an argument is passed that is not a number fail immediately.
Previously the check was done after the baur repository was located and the
database connection was established. Both steps are unnecessary in this case.

ls inputs: correct usage string

The usage output of ls inputs was wrong:
- APP_DIR was missing as supported arguments,
- it showed 2x the same argument but it is only supported to pass it 1x

init bashcomp: print how to load completions in current sessions

After writing the completion script to a file, print which command can be run to
use the completion script in the current shell session.

completion: limit "init repo" completion to directory names

"init repo" only supports directory names as arguments.
Limit the generated shell completion options to directory names.

completion: disable file completion for commands not supporting path arguments

Disabling shell completion for argument to filenames and directory for commands
that do not support paths as arguments.

completion: complete "baur status" arguments

Add shell completion support for the arguments of "baur status".
The supported arguments are the same then for "baur run". Therefore the same
completion function is used.

completion: complete "baur run" arguments

Add a shell completion function that completes argument to TARGET or relative
directory names.

Completions are generated for all possible completion which can become a lot.

The following completion options are generated for the baur-example repository,
when in the root directory 'baur run<TAB>' is entered in bash.

    a-team/go/src/github.com/simplesurance/hello-server  rng
    a-team/myredis                                       rng.*
    hello-server                                         *.rng.build
    hello-server.*                                       rng.build
    *.hello-server.build                                 *.rng.check
    hello-server.build                                   rng.check
    myredis                                              *.rng.test
    myredis.*                                            rng.test
    *.myredis.build                                      unixtime
    myredis.build                                        unixtime.*
    project-pink-einstein/rng                            *.unixtime.build
    project-pink-einstein/unixtime                       unixtime.build

completion: complete "ls apps" arguments

Add a custom shell completion function for "ls apps" arguments.
When the current directory is part of a baur repository, the argument is
completed to app-names and relative app directories.
The completed app directories are relative to the current working directory.

The following arguments are suggested by bash when "baur ls apps <tab>" is
entered in the root of the baur-examples git:
    a-team/go/src/github.com/simplesurance/hello-server  project-pink-einstein/rng
    a-team/myredis                                       project-pink-einstein/unixtime
    hello-server                                         rng
    myredis                                              unixtime

```

### DIFF
--- a/internal/command/completion.go
+++ b/internal/command/completion.go
@@ -17,7 +17,7 @@ func mockGitCommitID() (string, error) {
 func completeAppNameAndAppDir(
 	_ *cobra.Command,
 	_ []string,
-	toComplete string,
+	_ string,
 ) ([]string, cobra.ShellCompDirective) {
 	repo, err := findRepository()
 	if err != nil {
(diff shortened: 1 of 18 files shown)


internal/command/completion.go
```diff
@@ -0,0 +1,120 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/spf13/cobra"
+
+	"github.com/simplesurance/baur/v3/internal/log"
+	"github.com/simplesurance/baur/v3/pkg/baur"
+)
+
+func mockGitCommitID() (string, error) {
```
FranciscoKurpiel, line 13: I don´t know about the naming here. What about `noGitCommit`, or something like that?

FranciscoKurpiel (approved):

internal/command/completion.go
fho: yes, I also don't like the current naming much, `noGitCommit` is imho confusing

PR #419, opened 2023-02-01
Update to use doublestar's WithFilesOnly instead of our own filter

### LINKED ISSUE #418: fileglob resolver: use WithFilesOnly option for doublestar.FilepathGlob
doublestar supports a new option to only return files as results.
Use that option instead of filtering out all dirs on our own fs/fileglob.go

https://github.com/bmatcuk/doublestar/releases/tag/v4.5.0

prachurjya15:
Closes #418


fho (approved):

fho:
thank you @prachurjya15

PR #502, opened 2024-01-15
inputfiles: use git object IDs as input file content digests

fho:
If the baur repository is also a git repository and the git binary is available,
baur now uses git object IDs as file content digest.
The object IDs for all tracked files are read from the git repository.
If an inputfile is tracked and unmodified, the existing git object ID can be
used.
This does not work for files that are part of a git submodule.
Depending on the number of input files and hardware this can be much faster.
It saves CPU time because the digest does not need be
calculated and disk I/O because all input files don't have to be read anymore.
Only "git ls-files" needs to be executed and it's output parsed.
The digest of an InputFile is still a SHA384. It is the hash of the aggregation
of the git object id of the content and the SHA384 of it's relative file path.

If an file is not tracked, tracked but modified or part of a git submodule,
 "git hash-object" is executed to calculate the object ID. 
This ensures that the stored baur object ID for the
file is the same, if it is later added to the repository and the task is rerun.

Executing "git hash-object" is slower then calculating the sha384 digest in baur
itself. The typical baur usecase is that it runs in CI for a checked out git
repository, where all files are tracked. Thereore it is neglegtible.
This can be improved in a follow-up by calculating the object ID in baur without
running an external command.

If the baur repository is not a git repository, file digests are calculated as
before.

### DIFF
--- a/internal/vcs/git/files.go
+++ b/internal/vcs/git/files.go
@@ -1,0 +1,190 @@
+package git
+
+import (
+	"bufio"
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+
+	"github.com/simplesurance/baur/v3/internal/exec"
+)
+
+const (
+	// source of the descriptions: git-ls-files manpage (Git 2.43.0)
+
+	// H tracked file that is not either unmerged or skip-worktree
+	ObjectStatusCached byte = 'H'
+	// S tracked file that is skip-worktree
+	ObjectStatusSkipWorktree byte = 'S'
+	// M tracked file that is unmerged
+	ObjectStatusUnmerged byte = 'M'
+	// R tracked file with unstaged removal/deletion
+	ObjectStatusRemoved byte = 'R'
+	// C tracked file with unstaged modification/change
+	ObjectStatusChanged byte = 'C'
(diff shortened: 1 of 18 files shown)


internal/vcs/git/files.go
```diff
@@ -0,0 +1,191 @@
...
+	return 0, nil, nil
+}
+
+// LsFiles retrieves information about files in the index and working tree.
+// The information is sent to ch.
+// For a path multiple objects might be sent, if the status is in index and working tree differs.
+// The function closes ch when it terminates.
+func LsFiles(ctx context.Context, dir string, skipUntracked bool, ch chan<- *Object) error {
+	defer close(ch)
+	// TODO: test with paths containing spaces
```
nocive, line 82: shouldn't it "just work" ?

nocive (approved):

nocive:
@FranciscoKurpiel please review

internal/vcs/git/files.go
fho: Thanks, I forgot about that.

Yes, if the output is parsed correctly. If -z is not passed to git ls-files whitespaces and other special chars will also appear quoted.
I changed an existing  testcase to use a filename with whitespaces and utf8 chars to test it.

PR #566, opened 2024-05-22
support creating and showing information about releases

fho:
This PR introduces a new concept of "releases" to baur.
Releases are named snapshot of task runs.
Optionally user-specified metadata (arbitrary data) can be attached to a
release. Because it is stored in the database, it should not be too huge.

Releases provides a static view of task runs, including their
outputs. This allows to query artifact URIs without having to ensure that the
environment in that baur is run is exactly the same, to calculate the same total
input digests, to find the existing task runs.
Release information can even be queried without having access to the git
repository containing the baur configuration files, by specifying the URL to the
PostgreSQL server via an environment variable.
It is also much faster because input files do not have to resolved and digests
not calculated.

Information about a release are shown in JSON format, it includes the URIs of
all outputs that the tasks created. Attached metadata can either be written to a
separate file or shown as part of the JSON output, base64 encoded.

See commit messages for more details


internal/command/helpers.go
```diff
@@ -25,6 +25,12 @@ type Formatter interface {
 	Flush() error
 }
 
+var ErrPSQLURIMissing = errors.New(
+	"PostgreSQL connection information is missing.\n" +
+		"- set postgres_url in your repository config or\n" +
+		"- set the $" + envVarPSQLURL + "environment variable",
```
nocive, line 31:
```suggestion
		"- set the " + envVarPSQLURL + " environment variable",
```

internal/command/release_create.go
```diff
@@ -0,0 +1,181 @@
...
+Because it is stored in database, this should not consume too much disk space.
+
+By default all tasks are included in the release. This can be narrowed via the
+--include flag. See %s for more information about the TARGET syntax.
+
+Exit Codes:
+  0 - Success
+  1 - Error
+  %d - Release with the same name already exist
+  %d - One or more tasks are in pending state
```
nocive, line 40: Don't you wish to document the "not exists" exit code here as well?

```
exitCodeAlreadyExist     = 2
exitCodeTaskRunIsPending = 3
exitCodeNotExist         = 4
```

internal/command/release_create.go
```diff
@@ -0,0 +1,181 @@
...
+
+func newReleaseCreateCmd() *releaseCreateCmd {
+	cmd := releaseCreateCmd{
+		Command: cobra.Command{
+			Use:               "create NAME",
+			Short:             "create a release",
+			Long:              strings.TrimSpace(releaseCreateLongHelp),
+			Args:              cobra.ExactArgs(1),
+			Example:           strings.TrimSpace(releaseCreateExample),
+			ValidArgsFunction: nil, // FIXME: implement completion
```
nocive, line 65:
```suggestion
			ValidArgsFunction: nil, // TODO: implement completion
```

internal/command/release_create.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	cmd.Flags().BoolVarP(&cmd.requireCleanGitWorktree, flagNameRequireCleanGitWorktree, "c", false,
+		"fail if the git repository contains modified or untracked files")
+	cmd.Flags().StringVarP(&cmd.metadataFile, "metadata", "m", "",
+		"path to a file containing additional data that is stored with the release",
+	)
+	cmd.Flags().StringSliceVar(&cmd.includes, "include", nil,
+		"tasks to include in the release, by default all are included,\n"+
+			"supports the same TARGET syntax then 'baur run'")
```
nocive, line 76:
```suggestion
			"supports the same TARGET syntax as 'baur run'")
```

internal/command/release_create.go
```diff
@@ -0,0 +1,181 @@
...
+		exitOnErrf(err, "opening metadata file failed")
+
+		defer metadataReader.Close()
+
+		release.Metadata = metadataReader
+	}
+
+	err = storageClt.CreateRelease(ctx, &release)
+	if errors.Is(err, storage.ErrExists) {
+		stderr.PrintErrf("release with name %q already exist, release names must be unique", releaseName)
```
nocive, line 128:
```suggestion
		stderr.PrintErrf("release with name %q already exists, release names must be unique", releaseName)
```

internal/command/release_create.go
```diff
@@ -0,0 +1,181 @@
...
+		exitFunc(exitCodeAlreadyExist)
+	}
+	exitOnErr(err, "storing release information in database failed")
```
nocive, line 131:
```suggestion
	exitOnErr(err, "failed storing release information in database")
```

internal/command/upgrade_db.go
```diff
@@ -29,9 +37,10 @@ type upgradeDbCmd struct {
 func newUpgradeDatabaseCmd() *upgradeDbCmd {
 	cmd := upgradeDbCmd{
 		Command: cobra.Command{
-			Use:               "db",
+			Use:               "db [POSTGRES_URL]",
```
nocive, line 40: shouldn't this be using `envVarPSQLURL` too?

pkg/storage/postgres/insert_test.go
```diff
@@ -119,3 +94,63 @@ func TestSaveTaskRun(t *testing.T) {
...
+func TestCreateReleaseUserData(t *testing.T) {
+	client, cleanupFn := newTestClient(t)
+	defer cleanupFn()
+
+	require.NoError(t, client.Init(ctx))
+
+	runID, err := client.SaveTaskRun(ctx, newTaskRunWithoutOutputs())
+	require.NoError(t, err)
+
+	data := ` DATA€€¹³¹²æſ¢„ĸłʒ³\²😊æſðł¢„ł„¢łðđ	`
```
nocive, line 149: :stuck_out_tongue_winking_eye:

nocive (commented):
Looks great, the commands names and arguments seem intuitive to me and currently I don't see the need to have more options or flags.

If you want more feedback on specific parts let me know.

internal/command/release_create.go
fho: FIXME actually triggers the CI check to fail, so I won't forget it before merging it,
TODO does not

internal/command/upgrade_db.go
fho: It supports both, the env var and passing the URL as parameter, same "baur init db"

internal/command/release_create.go
fho: `release create` does not exit with `exitCodeNotExist`

internal/command/release_create.go
```diff
@@ -0,0 +1,191 @@
...
+		if len(c.includes) != 0 {
+			fatalf("could not find any tasks matching %s", strings.Join(c.includes, ","))
+		}
+		fatalf("could not find any tasks in the baur repository")
+	}
+
+	storageClt := mustNewCompatibleStorageRepo(repo)
+	runIDs := c.fetchTaskIDs(ctx, repo, storageClt, tasks)
+
+	release := storage.Release{Name: releaseName, TaskRunIDs: runIDs}
```
FranciscoKurpiel, line 125: - Shouldn't the release have more information associated to it, like commit id and timestamp?
- Do the release enforce having only runs from the same commit id? Shouldn't we enforce that?

FranciscoKurpiel (approved):
Nice improvement! The command names look fine to me too.

internal/command/release_create.go
fho: > Shouldn't the release have more information associated to it, like commit id

I'm undecided if the release commit id should be stored by baur, it can also part of the user-defined metadata.

>  and timestamp?

Yes, I probably will have to add a release creation timestamp, to be able to introduce a command to delete all releases older than X.

FranciscoKurpiel: > I'm undecided if the release commit id should be stored by baur

One alternative idea would be to always have a git tag with the same name as the baur release. I can't imagine a case where this wouldn't be desirable.

internal/command/release_create.go
fho: will be fixed in https://github.com/simplesurance/baur/pull/570

internal/command/release_create.go
fho: will be fixed in https://github.com/simplesurance/baur/pull/570

internal/command/release_create.go
fho: discussed face2face last week :-)

PR #570, opened 2024-05-28
add "release create" command

fho:
Introduce a "release create" command.
It creates a named snapshot of the state of a list of tasks in EXIST state.
Optionally user-specified metadata (arbitrary data) can be stored with the
release. Metadata is read from a file and stored in the database.

Releases will provide a static view of baur runs, including their
outputs, without having to ensure that the environment must be the same
when the outputs have been created.
When "baur status" + "baur ls outputs" is used there can be various
factors that make the result hard to reproduce, e.g:
- outputs might have been overwritten for the same input files via
  "baur run --force",
- change of the used tooling (go list) or  different environment
  variables used as inputs can result in different total input digests

In follow up further commands are added to interact with releases.

### DIFF
--- a/internal/command/release_create.go
+++ b/internal/command/release_create.go
@@ -85,7 +85,7 @@ func newReleaseCreateCmd() *releaseCreateCmd {
 	const includeFlagName = "include"
 	cmd.Flags().StringArrayVar(&cmd.includes, includeFlagName, nil,
 		"Target to be included in the release, by default all are included.\n"+
-			"It supports the same TARGET syntax then 'baur run',\n"+
+			"It supports the same TARGET syntax as 'baur run',\n"+
 			"the flag can be specified multiple times")
 	_ = cmd.RegisterFlagCompletionFunc(
 		includeFlagName,
@@ -145,7 +145,7 @@ func (c *releaseCreateCmd) run(cmd *cobra.Command, args []string) {
 
 	err = storageClt.CreateRelease(ctx, releaseName, runIDs, metadataReader)
 	if errors.Is(err, storage.ErrExists) {
-		stderr.PrintErrf("release with name %q already exist, release names must be unique\n", releaseName)
+		stderr.PrintErrf("release with name %q already exists, release names must be unique\n", releaseName)
 		exitFunc(exitCodeAlreadyExist)
 	}
 	exitOnErr(err, "storing release information in database failed")
--- a/internal/command/release_create.go
+++ b/internal/command/release_create.go
@@ -59,7 +59,7 @@ type releaseCreateCmd struct {
 }
 
 func init() {
-	releaseCMd.AddCommand(&newReleaseCreateCmd().Command)
+	releaseCmd.AddCommand(&newReleaseCreateCmd().Command)
 }
 
 func newReleaseCreateCmd() *releaseCreateCmd {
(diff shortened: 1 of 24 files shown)


internal/command/release_create.go
```diff
@@ -0,0 +1,200 @@
...
+			Args:    cobra.ExactArgs(1),
+			Example: strings.TrimSpace(releaseCreateExample),
+			ValidArgsFunction: func(_ *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+				return nil, cobra.ShellCompDirectiveNoFileComp
+			},
+		},
+	}
+
+	cmd.Flags().BoolVarP(&cmd.requireCleanGitWorktree, flagNameRequireCleanGitWorktree, "c", false,
+		"fail if the git repository contains modified or untracked files")
```
nocive, line 80:
```suggestion
		"Fail if the git repository contains modified or untracked files.")
```

internal/command/release_create.go
```diff
@@ -0,0 +1,200 @@
...
+	cmd.Flags().StringVarP(&cmd.metadataFile, "metadata", "m", "",
+		"path to a file containing additional data that is stored with the release",
```
nocive, line 82:
```suggestion
		"Path to a file containing additional data that is stored with the release.",
```

internal/command/release_create.go
```diff
@@ -0,0 +1,200 @@
...
+	)
+
+	const includeFlagName = "include"
+	cmd.Flags().StringArrayVar(&cmd.includes, includeFlagName, nil,
+		"Target to be included in the release, by default all are included.\n"+
+			"It supports the same TARGET syntax as 'baur run',\n"+
+			"the flag can be specified multiple times")
```
nocive, line 89:
```suggestion
			"the flag can be specified multiple times.")
```

nocive (approved):
lgtm :star:

internal/command/release_create.go
fho: all flag descriptions start with a lowercase, this is consistent with how gnu tools are doing it(cp, ls, cat, etch)

internal/command/release_create.go
fho: Most flag descriptions aren't whole sentences and do not end in a ".", this one is though.
Not sure if this should then also end up in a "." or be more consistent with the other flag descriptions and not have one at the end

internal/command/release_create.go
```diff
@@ -0,0 +1,200 @@
...
+
+	cmd.Flags().BoolVarP(&cmd.requireCleanGitWorktree, flagNameRequireCleanGitWorktree, "c", false,
+		"fail if the git repository contains modified or untracked files")
+	cmd.Flags().StringVarP(&cmd.metadataFile, "metadata", "m", "",
+		"path to a file containing additional data that is stored with the release",
+	)
+
+	const includeFlagName = "include"
+	cmd.Flags().StringArrayVar(&cmd.includes, includeFlagName, nil,
+		"Target to be included in the release, by default all are included.\n"+
```
nocive, line 87: This starts with uppercase and has complete sentences.

internal/command/release_create.go
nocive: Yeah, the other comments were in line with what this flag had.

fho: I shortened the description to not being full sentences anymore, new one:
```
      --include stringArray          tasks matching a TARGET string to include in the release,
                                     TARGET has the same syntax as used in 'baur run',
                                     the flag can be specified multiple times
```

internal/command/release_create.go
fho: fixed, thx

internal/command/release_create_test.go
```diff
@@ -0,0 +1,63 @@
+//go:build dbtest
+// +build dbtest
```
fho, line 2:
```suggestion
```

nocive (approved):

fho:
PR was rebased + merged manually

PR #571, opened 2024-05-28
add "release exists" command

fho:
Add the command "release exists". It checks if a release with give name
exists.

It can be used without a baur repository by setting the
BAUR_POSTGRESQL_URL environment variable.

### DIFF
--- a/internal/command/helpers.go
+++ b/internal/command/helpers.go
@@ -25,6 +25,12 @@ type Formatter interface {
 	Flush() error
 }
 
+var ErrPSQLURIMissing = errors.New(
+	"PostgreSQL connection information is missing.\n" +
+		"- set postgres_url in your repository config or\n" +
+		"- set the " + envVarPSQLURL + "environment variable",
+)
+
 var targetHelp = fmt.Sprintf(`%s is in the format %s
 Examples:
 - 'shop' matches all tasks of the app named shop
@@ -111,13 +117,8 @@ func mustArgToApp(repo *baur.Repository, arg string) *baur.App {
 func newStorageClient(psqlURI string) (storage.Storer, error) {
 	uri := psqlURI
 
-	if envURI := os.Getenv(envVarPSQLURL); len(envURI) != 0 {
-		log.Debugf("using postgresql connection URL from $%s environment variable",
-			envVarPSQLURL)
-
+	if envURI := getPSQLURIEnv(); envURI != "" {
 		uri = envURI
-	} else {
-		log.Debugf("environment variable $%s not set", envVarPSQLURL)
 	}
 
 	var logger postgres.Logger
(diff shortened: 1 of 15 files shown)


internal/command/helpers.go
```diff
@@ -129,32 +130,71 @@ func newStorageClient(psqlURI string) (storage.Storer, error) {
...
+	exitFunc(exitCodeError)
+	return ""
 }
 
-func getPSQLURI(cfg *cfg.Repository) string {
-	if url := os.Getenv(envVarPSQLURL); url != "" {
-		return url
+func getPSQLURIEnv() string {
+	if envURI := os.Getenv(envVarPSQLURL); len(envURI) != 0 {
+		log.Debugf("using postgresql connection URL from $%s environment variable",
```
nocive, line 152: Env variable references in documentation don't usually contain the dollar sign.
```suggestion
		log.Debugf("using postgresql connection URL from %s environment variable",
```

internal/command/helpers.go
```diff
@@ -129,32 +130,71 @@ func newStorageClient(psqlURI string) (storage.Storer, error) {
...
+			envVarPSQLURL)
+
+		return envURI
+	}
+
+	log.Debugf("environment variable $%s not set", envVarPSQLURL)
```
nocive, line 158:
```suggestion
	log.Debugf("environment variable %s not set", envVarPSQLURL)
```

pkg/storage/postgres/query.go
```diff
@@ -465,3 +465,19 @@ func (c *Client) TaskRuns(
 
 	return nil
 }
+
+func (c *Client) ReleaseExists(ctx context.Context, name string) (bool, error) {
+	const query = `
```
nocive, line 470: Not using `COUNT` might perform better:

```
SELECT EXISTS(SELECT 1 FROM release WHERE name = $1)
```

nocive (approved):

pkg/storage/postgres/query.go
fho: The current query also has `LIMIT 1`.
I expect that the difference will be barely measurable, seems not worth to me to measure, compare them + eventually adapt the code.

nocive: LIMIT is notoriously bad at optimization, it also doesn't make sense to use LIMIT here since `name` should be unique. Returning more than 1 row should be detected as an error.

internal/command/helpers.go
fho: fixed

internal/command/helpers.go
fho: fixed

fho:
merged manually

internal/command/helpers.go
```diff
@@ -129,32 +130,71 @@ func newStorageClient(psqlURI string) (storage.Storer, error) {
...
+	if url := getPSQLURIEnv(); url != "" {
+		return url, nil
 	}
 
-	return cfg.Database.PGSQLURL
+	repo, err := findRepository()
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return "", fmt.Errorf("can not locate postgresql database\n"+
+				"- the environment variable $%s is not set\n"+
```
nocive, line 178: there's one here too :)

PR #572, opened 2024-05-30
add release download command

fho:
Add  "release download" command.
It downloads S3 outputs from task-runs that are associated with the
release. If an output was uploaded to multiple S3 destinations, only one
of them is downloaded.
By default all outputs of a release are downloaded. It is possible to
only download outputs for certain tasks, by specifying their tasks IDs
via the --tasks parameter. If a specified task is not part of the
release, or has no outputs that have been uploaded to S3 the command
will fail.
The command currently only supports plain text output. In a follow-up
support for JSON output can be added, containing all outputs and their
download paths.

To test the new command, support for starting a adobe/s3mock docker
container is added. The TestReleaseDownload test uploads and downloads
s3 artifacts to the test container.
It is also started in CI when running tests on Linux.
The test is disabled when running the tests on Windows.


nocive (commented):
lgtm :smiling_face_with_tear:

nocive:
```
"%s: %s: skipping download, output already has been downloaded from another destination \n",
```

> output has already been downloaded

PR #582, opened 2024-06-10
upgrade cfg: support upgrading configuration files to v7

fho:
```
upgrade cfg: support upgrading configs to version 7

The configuration files are downwards compatible since version 5, only
new fields have been added.
The Repository configuration file is the same since v5, only the
cfg_version needs to be increased.

Implement upgrading v5 and v6 baur config files to v7, by loading the
repository config file and changing the cfg_version value to 7.

cfg: remove baur/v2 pkg dependency

For being able to upgrade the baur cfg from version 5 to 6, the cfg
package from baur/v2 was imported.
The Repository configuration file object has not changed since since cfg
version 5, only the cfg_version value has increased.
Use the cfg package from the current version, to remove the unnecessary
dependency to the baur/v2 package.

baur: fix: baur.AppCfgFile imported from another baur version

The Loader accidentally imported baur.AppCfgFile from the baur v0
package instead of using the constant from the baur package of the same
version.

upgrade cfg: remove support for upgrading from v4->v5

Config version 4 was introduced more then 2 years, remove support for
migrating from it via "upgrade cfg".

cfg: increase version number to 7

upgrade configs: fix: wrong version nr in usage

```


fho:
merged manually

PR #588, opened 2024-06-20
add cleanup db command

fho:
Add a "cleanup db" command.
The commands removes all tasks runs that have been started before a
given timestamp.
When run it prints what it will do and gives the user 5sec to abort.
This can be skipped by passing the --force parameter.
It prints the number of deleted records.

--pretend can be passed to not do anything and only print the number of
records that would be deleted.


fho:
merged manually

PR #589, opened 2024-06-21
support cleaning up old releases

fho:
```
support cleaning up old releases

- change the "db cleanup" command syntax to accept a positional argument
  to specify if releases or tasks should be deleted,
- rename the --task-runs-before parameter to --before
- implement ad DeletedReleases method for the psql storage client, that
  works analogous to DeletedTaskRuns

add created_at timestamp to release table

This will allow to cleanup releases that are older then X.

For all existing releases the created_at timestamp is set to the current
timestamp.
```

### DIFF
--- a/internal/command/cleanup_db.go
+++ b/internal/command/cleanup_db.go
@@ -1,31 +1,37 @@
 package command
 
 import (
+	"context"
 	"fmt"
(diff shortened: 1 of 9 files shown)


fho:
merged manually

PR #627, opened 2025-02-08
Disallow .. in application_dirs to prevent unintended directory resolution

viniciusartur:
Disallow `..` and `../` in **application_dirs** to prevent unintended directory resolution. This avoids configs pointing outside the working directory, which could lead to unexpected behavior in builds and absolute path leaks in errors.

New error message:
`
ERROR: validating repository config "/home/user/project/.baur.toml" failed: Discover.application_dirs: contains an invalid path: parent directory access ('..') is not allowed
`

### DIFF
--- a/pkg/cfg/repository.go
+++ b/pkg/cfg/repository.go
@@ -3,6 +3,8 @@ package cfg
 import (
 	"fmt"
 	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/pelletier/go-toml"
 )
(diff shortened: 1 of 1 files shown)


fho:
Hello @viniciusartur,
thanks a lot for your PR.

The validation makes sense but checking if the path contains ".." only catches a subset of the cases.
It also disallows using ".." despite using it is totally fine, as long as the path is still part of the repository directory.

To make the validation more resilient it should verify that the application directory is a children of the repository directory, after resolving symlinks.

The repository directory path is not available in `Discover.validate`.
But the validation could also be done after the configuration file was parsed, in pkg/baur/loader.go.

viniciusartur:
Hello @fho,
Makes sense. I had a wrong key assumption that `.baur.toml` should be in the root of the repository.
It's hard for me to make this change in other point, because I'm quite fresh to this repository, so feel free to close this pull request.

fho:
> I had a wrong key assumption that `.baur.toml` should be in the root of the repository. 

Your assumption was correct. The baur repository root is where the `.baur.toml` file is.

PR #640, opened 2025-04-20
ci: update golangci-lint to version 2.1

fho:


fho:
> Nice! I was struggling to convert the local package option.

`golangci-lint migrate` did it :-)

Issue #642, opened 2025-04-22
gosource resolver: discover go.mod files


fho:
Change the gosource resolver to discover and return the paths of the involved go.mod files.

The user won't have to list the go.mod file manually anymore for go projects.
Tracking the go.mod file allows to detect that a task needs to be run when the specified version of a non-vendored 3.-party package or a minor option like the min. required go version changed.

PR #647, opened 2025-05-05
gosource: return path to go.mod files

fho:
The gosource resolver now returns the path to the go.mod as part of the result.
The go.mod file defines, among others the imported third-party modules and parameters for the build environment.

Including it in the gosource resolver results means, user do not need define the go.mod file manually as part of baur tasks anymore.

Closes: #642


nocive (approved):
:tada:

